Issue #4, opened 2020-02-21
Could support genesis transactions at URL


sklump:
https://github.com/andrewwhitehead/indy-vdr/blob/414b1903c2a9efb99ee2a9e6c844394f822ca094/indy-vdr-proxy/src/main.rs#L52

Could support URL here?

Any issues I put up here are tiny in the face of how awesome this software is as a whole.

andrewwhitehead:
It would be easier to support there in the proxy because it already has an HTTP client library. I didn't really want to add that as a dependency on the main rust library. Also thanks!

PR #11, opened 2020-02-28
Add support for downloading genesis transaction files

andrewwhitehead:
Addresses #4

### DIFF
--- a/indy-vdr-proxy/src/main.rs
+++ b/indy-vdr-proxy/src/main.rs
@@ -14,11 +14,17 @@ use std::process::exit;
 use std::rc::Rc;
 use std::time::SystemTime;
 
+#[cfg(feature = "fetch")]
+use hyper::body::Buf;
 use hyper::service::{make_service_fn, service_fn};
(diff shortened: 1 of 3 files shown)


indy-vdr-proxy/src/main.rs
```diff
@@ -48,8 +54,49 @@ pub struct AppState {
...
 
+#[cfg(feature = "fetch")]
+async fn fetch_transactions(genesis: String) -> VdrResult<PoolTransactions> {
+    let https = HttpsConnector::new();
+    let client = Client::builder().build::<_, hyper::Body>(https);
+    let mut res = client
+        .get(
+            genesis
+                .parse()
+                .with_err_msg(VdrErrorKind::Config, "Error parsing genesis URL")?,
```
sklump: Could put the genesis URL itself into the error message, so the troubleshooter can identify what's wrong with it? (e.g.,
```
.with_err_msg(VdrErrorKind::Config, format!("Error parsing genesis URL {}", genesis.as_str())
```

PR #13, opened 2020-03-05
IS-1501:Libindy use Indy VDR Part 2

Artemkaaas:
Signed-off-by: artem.ivanov <artem.ivanov@dsr-company.com>

### DIFF
--- a/libindy_vdr/src/ledger/requests/cred_def.rs
+++ b/libindy_vdr/src/ledger/requests/cred_def.rs
@@ -53,14 +53,6 @@ impl CredentialDefinition {
     }
 }
 
-impl From<CredentialDefinition> for CredentialDefinitionV1 {
-    fn from(cred_def: CredentialDefinition) -> Self {
-        match cred_def {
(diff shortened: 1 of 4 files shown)


libindy_vdr/src/ledger/requests/cred_def.rs
```diff
@@ -53,6 +53,14 @@ impl CredentialDefinition {
...
 }
 
+impl From<CredentialDefinition> for CredentialDefinitionV1 {
+    fn from(cred_def: CredentialDefinition) -> Self {
+        match cred_def {
+            CredentialDefinition::CredentialDefinitionV1(cred_def) => cred_def
+        }
+    }
+}
+
```
andrewwhitehead: I know this is used a lot in the anoncreds code but it's a bad pattern which is why I removed it. In indy-credx it does the downcasting in each place that the type is used. I can merge it if that's needed though.

Artemkaaas: I agree. I'll do the same for Libindy.

PR #25, opened 2020-03-17
Update proxy response format, add pool auto-refresh

### LINKED ISSUE #16: Clean up response output from proxy server
- Set application/json header
- Return node info in a header so that the body is one JSON response
- Return formatted text when text/html is requested

Requires some refactoring to support alternative result formatters.

### LINKED ISSUE #17: Enable auto-refresh of validators in proxy server
Should perform refresh on a timer, with a command line flag to adjust or disable the interval.

andrewwhitehead:
- Output HTML when text/html is requested or `?html` query string is used, otherwise output application/json and return timing information as a header (fixes #16)
- Add auto-refresh of validator pool. New `--refresh-interval` option, defaults to 120 minutes (fixes #17)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Indy-VDR
+# Indy-VDR (Verifiable Data Registry)
 
 [<img src="https://raw.githubusercontent.com/hyperledger/indy-node/master/collateral/logos/indy-logo.png" width="50%" height="auto">](https://github.com/hyperledger/indy-sdk/)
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -14,7 +14,7 @@
 
 ## Introduction
 
-This library is derived from [Hyperledger Indy SDK](https://github.com/hyperledger/indy-sdk) for the more limited use case of connecting to an Indy Node blockchain ledger, a Verifiable Data Registry. It is written in Rust and currently includes a Python wrapper and a standalone proxy server.
+This library is derived from [Hyperledger Indy SDK](https://github.com/hyperledger/indy-sdk) for the more limited use case of connecting to an [Indy Node](https://github.com/hyperledger/indy-node) blockchain ledger. It is written in Rust and currently includes a Python wrapper and a standalone proxy server.
```
swcurran: Should it also be mentioned here that there is an FFI interface?

andrewwhitehead: It could but that sounds like a separate issue ;)

PR #33, opened 2020-03-28
Clean up public API and rustdocs

andrewwhitehead:
- Avoid exporting internal methods and data types
- Add docstrings to public exports
- Add StateProofParser trait, state_proof_parser_fn method
- Rename a few methods for clarity
- Moved functions to `vdr_shared` crate - will be replaced with `indy-shared-rs` components

Hopefully this doesn't cause too many issues for the indy-sdk integration.
Run `cargo +nightly doc --no-deps --lib` to generate crate docs in target/doc


libindy_vdr/src/state_proof/types.rs
```diff
@@ -1,31 +1,24 @@
-/**
- Single item to verification:
- - SP Trie with RootHash
- - BLS MS
- - set of key-value to verify
-*/
+/// A single result from a state proof verification
```
jovfer, line 1: I think this structure is rather input for state proof verification parsed from incoming JSON. But it is not a final result yet

libindy_vdr/Cargo.toml
```diff
@@ -18,12 +18,12 @@ ffi = ["ffi-support", "logger"]
 logger = ["env_logger", "log"]
 zmq_dynamic = ["zmq"]
 zmq_vendored = ["zmq/vendored"]
-default = ["ffi", "log", "zmq_vendored"]
 local_nodes_pool = []
+rich_schema = []
```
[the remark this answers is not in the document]
andrewwhitehead: I've now enabled the data types when under `cfg(test)` but the integration tests will require the `rich_schema` feature to be manually enabled, I can't see a way around that.

libindy_vdr/src/config/mod.rs
```diff
@@ -1,4 +1,7 @@
-pub mod constants;
-pub mod types;
+pub(crate) mod constants;
+pub(crate) mod types;
 
-pub use types::{PoolConfig, ProtocolVersion};
+pub use types::PoolConfig;
+
+/// Library package version
+pub static LIB_VERSION: &str = env!("CARGO_PKG_VERSION");
```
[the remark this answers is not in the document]
andrewwhitehead: I don't understand the issue.

libindy_vdr/src/state_proof/types.rs
andrewwhitehead: Updated

vdr-shared/src/qualifier.rs
```diff
@@ -52,7 +56,7 @@ pub trait Qualifiable: From<String> + std::ops::Deref<Target = String> + Validat
     fn default_method(&self, method: Option<&str>) -> Self {
         let (prev_method, rest) = self.split();
         match prev_method {
-            Some(_) => Self::from((*self).clone()),
+            Some(_) => Self::from(self.to_string()),
```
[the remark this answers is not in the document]
andrewwhitehead: The trait doesn't currently extend Clone but this method works without requiring it.

vdr-shared/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "vdr-shared"
+version = "0.1.0"
+authors = ["Hyperledger Indy Contributors <hyperledger-indy@lists.hyperledger.org>"]
+edition = "2018"
+
+description = "This is the official SDK for Hyperledger Indy (https://www.hyperledger.org/projects), which provides a distributed-ledger-based foundation for self-sovereign identity (https://sovrin.org). The major artifact of the SDK is a c-callable library."
```
brentzundel, line 7: Is indy-vdr intended to be the official SDK for hyperledger indy?

swcurran: For the capabilities it provides - the ledger interface.  So not "the" official SDK, but one of several for different components.

brentzundel: might be good to modify this description then

PR #38, opened 2020-06-20
Move demo file, add instructions to run python demo

Patrik-Stas:
- Python demo was failing because argument passed to `build_custom_request(test_req)` was missing some fields
- Added instructions to run demo
- Moved `test.py` into `demo` directory
- Use Sovrin Buildernet by default in demo

Signed-off-by: Patrik <patrik.stas@absa.africa>


Patrik-Stas:
@swcurran roger that! I've updated it to download from Sovrin github without using external dependency

swcurran:
Thanks -- looks good to me.  Will leave it to @andrewwhitehead to review/merge it.

andrewwhitehead:
Looks good, thanks!

PR #39, opened 2020-06-21
WIP: Add nodejs ffi boilerplate

Patrik-Stas:
Based off libvcx nodejs wrapper but using `ffi-napi` compatible with NodeJS 10+

Signed-off-by: Patrik <patrik.stas@absa.africa>


andrewwhitehead:
Thanks! It would be good to integrate with the system library finder rather than hardcoding paths.

Issue #45, opened 2020-09-05
Responses from /submit endpoint not parseable with indy sdk (parseGetCredDefResponse, etc)


msembinelli:
Hi All,

I am running Indy-vdr-proxy in a network where connecting directly to indy-sdk pool would be blocked (FW only allows HTTP/HTTPS egress, so ports 9700-9799 for Sovrin StagingNet are blocked).

The idea here, is that I leverage the indy-vdr-proxy client side to proxy out to ledger in a more favorable way with current security settings. In my client application, I am replacing indy Ledger/pool calls with calls to the proxy on the /submit endpoint. This seems to be working great as far as providing prepared messages, and getting responses out of the proxy.

The issue lies when I attempt to use the indy sdk to parse the message. For example:

**Prepared message sent to proxy /submit**
This is a get cred def ledger request
```
{"reqId":1599263022696669846,"identifier":"HAwm7Mvmsw3FGAeLLDwZtA","operation":{"type":"108","ref":137738,"signature_type":"CL","origin":"KuG6kY3i7RwmWV8B4HnWCs","tag":"testcadev9999123654"},"protocolVersion":2}
```

**Response from vdr proxy /submit**
```
{
  "data": {
    "primary": {
      "n": "...",
      "r": {
        "master_secret": "...",
        "name": "..."
      },
      "rctxt": "...",
      "s": "...",
      "z": "..."
    }
  },
  "identifier": "HAwm7Mvmsw3FGAeLLDwZtA",
  "origin": "KuG6kY3i7RwmWV8B4HnWCs",
  "ref": 137738,
  "reqId": 1599263022696670000,
  "seqNo": 137739,
  "signature_type": "CL",
  "state_proof": {
    "multi_signature": {
      "participants": [
        "trusted_you",
        "regioit01",
        "EBPI-validation-node",
        "NECValidator",
        "lab10",
        "DigiCert-Node",
        "ayanworks",
        "Swisscom",
        "Absa"
      ],
      "signature": "Qq1tHugcFCdzaMGbhcEUZGNAcJX6ef6UdSqnRUKWTvJW6U9TigzREznTpi5FFigHbKBjhaY6i14mZRjX88YtyHVaGC3GqWwVpVVPzuRH7xXeNsYVfDhppKxZLnuKkJqNsEjBBvHJsjicwQZjKWYooiZg5JnGGGpH4DXSYMJvnjmbNP",
      "value": {
        "ledger_id": 1,
[...]

PR #46, opened 2020-09-05
fix: Return full json body instead of just result

msembinelli:
Return full json body instead of just result for `/submit` (prepared requests)
Ensure that `op` and `result` exist to retain indy-sdk parse compatibility.

Fixes https://github.com/hyperledger/indy-vdr/issues/45


andrewwhitehead (approved):
Looks good to me.

PR #54, opened 2021-02-24
fix type mismatch between golang and C

kukgini:
Signed-off-by: kukgini <kukgini@gmail.com>


andrewwhitehead:
Thanks @kukgini – any idea why this changed? Is it because a newer Go version is used? I can't find the documentation for what `import "C"` is supposed to provide.

andrewwhitehead (approved):

PR #55, opened 2021-02-25
Build updates

andrewwhitehead:
- Update the Github workflow to build binaries and Python wheel packages for x86_64 platforms
- Fix indy-vdr-proxy build on Windows
- Minor updates to nodejs and python wrapper metadata


indy-vdr-proxy/src/app.rs
```diff
@@ -11,7 +12,7 @@ pub struct Config {
 }
 
 pub fn load_config() -> Result<Config, String> {
-    let matches = App::new("indy-vdr-proxy")
+    let mut app = App::new("indy-vdr-proxy")
```
ryjones, line 15: Check warning on line 15 in indy-vdr-proxy/src/app.rs

 GitHub Actions
/ Run Checks (windows-latest)

indy-vdr-proxy/src/app.rs#L15
variable does not need to be mutable

ryjones (approved):
fixing the type complaint is optional

Issue #62, opened 2021-07-13
Support ZMQ socks proxy


vitalrev:
Same as INDY-SDK Issue [2399]( #https://github.com/hyperledger/indy-sdk/issues/2399)
must be implemented/copied in INDY-VDR too

_By using INDY-VDR behind a corporate proxy, a socks proxy must be configured for ZMQ socket connections.
Zero MQ Socket supports this by setting the ZMQ_SOCKS_PROXY property, for Rust Wrapper is it set_socks_proxy function.
The problem is, that the INDY-VDR doesn't support the possibility to set and pass on the property to ZMQ._

vitalrev:
I am preparing a pull request

vitalrev:
pull request is merged

PR #63, opened 2021-08-02
Support setting ZMQ socks proxy

vitalrev:
implementation for issue #62


andrewwhitehead:
Hi @vitalrev – the patch looks good, but needs rustfmt to be run (`cargo fmt`), thanks!

andrewwhitehead (approved):

PR #66, opened 2021-09-21
Import sha3 directly instead of through ursa

andrewwhitehead:
Ursa 0.3.7 was released with a different and incompatible version of the dependency, so this fixes a build error as well.


swcurran:
How come you are doing this?  Something different/bad coming from Ursa?

andrewwhitehead:
There's a compilation error at the moment due to Ursa changing from sha3 0.8 to 0.9 (which uses a different version of the digest crate) in the new patch release.

Issue #74, opened 2022-01-19
did:indy API changes


domwoe:
I started thinking about the required API changes to support did:indy:

- `build_nym_request` needs to support an optional `diddocContent` parameter
- `build_get_nym_request` needs to support optional `seqNo`  (versionId) and `timestamp` (versionTime) parameters

Unfortunately, Rust does not support functions with optional arguments. Hence, I see the following options:

1. Break the existing API, i.e add the optional arguments to the functions. Consumers of the API will need to adapt their code when updating Indy VDR
2. Add new functions, e.g. `build_nym_request_v2` that include the additional optional parameters
3. Add a feature flag to either build Indy VDR with a legacy API or did:indy API

Did I miss anything? What are your thoughts?

domwoe:
We decided to just add the optional parameters.

PR #80, opened 2022-02-21
Changes for did:indy

domwoe:
This is a WiP pull request for changes relating to did:indy

- [x] Adds new properties to `NymOperation`  (`diddocContent`)
- [ ] Adds new properties to `GetNymOperation`  (`seqNo`, `timestamp`)
- [x] Adds resolver module to resolve ledger objects according to  did:indy spec

## Resolver
The resolver is initialized per pool/ledger ([example](https://github.com/IDunion/indy-did-resolver/blob/e2d7d1971006ecc6a124b0be1febd2fed9270f8e/indy-did-driver/src/main.rs#L49))

## Backward compatibility

As mentioned in #74, I don't really know what's the best way to handle backward compatibility.
I've implemented one possibility for the `NymOperation`, i.e. the `build_nym_request` function in particular.
I've introduced two new functions `build_nym_request_did_indy` and `build_nym_request_legacy`. The former accepts the additional, optional `diddoc_content` parameter. Depending on the feature flag `did_indy` (Changed this from `legacy`, because otherwise the test suite would fail), either  `build_nym_request_did_indy` or `build_nym_request_legacy` gets exposed as `build_nym_request`.
I'm totally not sure if this is a reasonable approach :)

## Proxy

Ideally, the proxy would expose the resolver as an  API compatible with the universal resolver. Currently, the proxy manages only a single pool/ledger. We'd need to change the proxy to manage multiple pools/ledgers.

## Wrapper

How do we change the wrappers with respect to backward compatibility?


domwoe:
Btw, maybe it would make sense to create a `did-indy` branch. What do you think?

c2bo:
+1 for the idea of a did-indy branch as a WIP branch while the spec and the implementations continue to evolve.

libindy_vdr/Cargo.toml
```diff
@@ -17,13 +17,15 @@ crate-type = ["staticlib", "rlib", "cdylib"]
...
 ffi = ["ffi-support", "logger"]
+did_indy = []
 logger = ["env_logger", "log"]
 zmq_vendored = ["zmq/vendored"]
 local_nodes_pool = []
 rich_schema = ["indy-data-types/rich_schema"]
 default = ["ffi", "log", "zmq_vendored"]
 
 [dependencies]
+chrono = "0.4.12"
```
andrewwhitehead, line 28: `chrono` is not supported anymore and we switched to `time`, see [commit here](https://github.com/hyperledger/indy-vdr/commit/d985f422b2d178b9bfcaa872bb0dff2f64a5c693)

andrewwhitehead:
I think I would prefer to just add `diddocContent` as a parameter to `build_nym_request` instead of having the feature flag and gating the method signature on that. If the DID doc content is provided then the request may not appear valid to non-did-indy servers, but otherwise nothing changes there. The change to the FFI method signature (an additional optional parameter) can be introduced in a minor version update.

domwoe:
Superseded by #84

PR #84, opened 2022-03-15
did:indy

domwoe:
Replaces #80 

## Support for additional optional fields in `nym` and `get_nym` requests
- `nym` request has optional `diddoc_content` field
- `get_nym` request has optional `seq_no` and `timestamp` fields
- Builder functions, ffi and wrappers are adapted accordingly
- Proxy allows `seq_no` and `timestamp` query params for `/nym` request

## Resolver
- Add resolver library with `PoolResolver` and `PoolRunnerResolver` (using callbacks for FFI)
- Add resolver to Python wrapper and test
- `PoolRunnerResolver` does not work with legacy resolution using attrib endpoint, since I can't get nested pool requests to work
- `PoolRunnerResolver` needs refactoring and proper error handling. I don't know how to handle/bubble up errors in closures  best

## Miscellaneous
- Removed `chrono` dependency in favor for `time-rs`


swcurran:
@domwoe -- decisions were made on the did:indy call between last week and this on the self-certifying DIDs, and it impacts the interface. You might want to update this to include that feature (or add it later).  The summary:

- an optional `version` field to be include in the DID publish interface with values:
  - `0` (default, no DID self-certification type specified)
  - `1` existing `did:sov` method
  - `2` new `did:indy` method.
- only included on a `CREATE` DID operation -- if included in an `UPDATE`, the ledger will reject the transaction
- if set to `1` or `2`, the ledger will check the DID/verkey relationship and if not correct, will reject the transaction

In future, a ledger MAY decide to enforce a minimum version for new DIDs.

@cjhowland @dbluhm -- flagging you to make sure I have it right.

domwoe:
@swcurran  I'll do. 
I'll restate a bit for the indy vdr perspective  to make sure I got it:
- The `NYM` tx and its builders will get an optional `version` field.  By default, this field is not set. Users will need to set this field proactively.
- The response of a `GET_NYM` tx will have an optional `version` field as well. It is only set if the respective `NYM` tx included it in the first place.

Concerning the resolver:

Do we want to define a field in the `didDocumentMetadata` to convey this information as well?

swcurran:
A couple of things:

- `The response of a GET_NYM tx will have an optional version field as well. It is only set if the respective NYM tx included it in the first place.`
  - AFAIK the GET_NYM returns from the ledger all of the "latest" data from a NYM, so that if you update a NYM, but for example, change only the verkey (a key rotation), the rest of the data (e.g. the `version` and `diddocContent`) will be picked up from the previous transactions for that NYM.  So `version` would always be returned.  Right @cjhowland / @dbluhm ?

We definitely want that data element returned from the resolver, in whatever way is best. didDocumentMetadata seems like a good way, but I'm not an expert on resolution.  If you are unsure, perhaps check with Markus S.?

swcurran:
I think in that case, the resolver would not include it.  That's based on my thought on checking the state proof, but that may not be relevant.

PR #85, opened 2022-03-17
Feature/general indy resolver

domwoe:
Builds upon #80 

- Adds a VDR API with a general DID resolver. Network configuration can be taken from a Github Repo or a local folder structure
- Initial support for multiple ledgers and DID resolution with indy-vdr-proxy

The VDR API uses Async/Await and is therefore not suited for FFI, but I'll look into it.

I think it would help to create a `did-indy` branch to have smaller PRs at some point :)

### DIFF
--- a/libindy_vdr/src/vdr/vdr.rs
+++ b/libindy_vdr/src/vdr/vdr.rs
@@ -475,21 +475,39 @@ impl RunnerVdr {
                                     //         }),
                                     //     );
                                 } else {
-                                    let diddoc = Some(doc.to_value().unwrap());
-                                    let md = if let Metadata::DidDocumentMetadata(md) = metadata {
-                                        Some(md)
-                                    } else {
-                                        None
-                                    };
+                                    // let diddoc = Some(doc.to_value().unwrap());
+                                    // let md = if let Metadata::DidDocumentMetadata(md) = metadata {
+                                    //     Some(md)
+                                    // } else {
+                                    //     None
+                                    // };
 
-                                    let result = ResolutionResult {
-                                        did_resolution_metadata: None,
-                                        did_document: diddoc,
-                                        did_document_metadata: md,
-                                    };
-                                    if !doc.diddoc_content.is_none() {
-                                        callback(Ok(serde_json::to_string_pretty(&result).unwrap()))
-                                    }
+                                    // let result = ResolutionResult {
+                                    //     did_resolution_metadata: None,
+                                    //     did_document: diddoc,
+                                    //     did_document_metadata: md,
+                                    // };
+                                    // if !doc.diddoc_content.is_none() {
+                                    //     callback(Ok(serde_json::to_string_pretty(&result).unwrap()))
+                                    // }
+                                }
+
+                                // For now, until if/else above is used
+
+                                let diddoc = Some(doc.to_value().unwrap());
+                                let md = if let Metadata::DidDocumentMetadata(md) = metadata {
+                                    Some(md)
+                                } else {
+                                    None
+                                };
+
+                                let result = ResolutionResult {
+                                    did_resolution_metadata: None,
+                                    did_document: diddoc,
+                                    did_document_metadata: md,
+                                };
+                                if !doc.diddoc_content.is_none() {
+                                    callback(Ok(serde_json::to_string_pretty(&result).unwrap()))
                                 }
                             }
                             _ => {}
--- a/libindy_vdr/src/vdr/vdr.rs
+++ b/libindy_vdr/src/vdr/vdr.rs
@@ -6,15 +6,15 @@ use std::fs;
 use std::path::PathBuf;
 
 use crate::common::error::prelude::*;
-use crate::ledger::responses::Endpoint;
+use crate::ledger::RequestBuilder;
 use crate::pool::genesis::PoolTransactions;
 use crate::pool::helpers::{perform_ledger_request, perform_refresh};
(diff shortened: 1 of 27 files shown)


andrewwhitehead:
`git2` is a pretty heavy dependency, which is bringing in openssl and breaking the tests at the moment (that part can probably be fixed by adding openssl-dev to the test containers). Can you put the git support behind a feature flag?

swcurran:
Could the git dependency be removed based and instead use HTTP and https://raw.githubusercontent.com?  It adds a requirement to the structure of the repo (e.g. paths must be predictable, branch must be known), but I think those are there regardless.  I'm assuming that an HTTP requirement would be lighter weight than a git one.

swcurran:
@domwoe -- I don't think the structure of the "indy-did-ledgers" (is that the name?) structure been formalized?  I don't recall that, but could be wrong.  I don't think that would be a big deal to propose something and have it discussed in the did:indy meetings.

domwoe:
There's a [section](https://hyperledger.github.io/indy-did-method/#dynamic-using-github).

Cloning the repo was the simplest since https://raw.githubusercontent.com/ does only work for files. If we don't want to use git, we could add a list of the networks at the root of the repo. Then, we could construct the raw content URLs to the genesis  txn files from this.

domwoe:
I tried to implement the FFI interface for the `VdrRunner` implementation. The `VdrRunner` is a wrapper around a collection of `PoolRunner`. However, I tried multiple ways to implement a pool refresh through that interface and can't get it working. Furthermore, I don't know how a client through FFI would get a pool handle to use the existing (pool) API. Otherwise, the entire existing API would need to be replicated in the `VdrRunner`.

Other issues in the FFI interface which I can't resolve are:
- I can't get the nested callback to `fetch_legacy_endpoint_with_runner` to work.
- Error handling in nested callbacks

Besides, one thing I'd like to add to this PR is to restructure the VDR API to allow passing pool configurations.

andrewwhitehead:
Would it make more sense to implement the multi-ledger features at the wrapper level? Then it could just handle multiple Pool instances behind the scenes, and things like git or http resolution of the configuration don't need to be built into the Rust component.

paulbastian:
I agree. Indy-vdr should be a simple core library that handles connectivity to "a" ledger. did:indy complexity should be handled best at indy-vdr-proxy or indy-did-driver

swcurran:
@domwoe -- what does that do to your implementation?  Obviously that would mean what was planned to go into Indy VDR needs to go into the various wrappers.  How can we get that work done?

domwoe:
@swcurran I think that's fine and that's what I thought until last week. 
I'll clean up a bit and try to implement the indy-vdr-proxy accordingly.  The indy-did-driver is implemented in this way already (See this [PR](https://github.com/IDunion/indy-did-resolver/pull/21)).

I might be able to look into the Python wrapper to make it more userfriendly, but only after finishing the rest and being sure that indy-vdr works with the new indy node implementation. The pool-specific resolver interface is already implemented in the Python wrapper.

I don't think I will be able to add functionality to the go and nodejs wrappers. 

@andrewwhitehead Could you help me a bit with the nested callbacks in the PoolRunner resolver?

andrewwhitehead:
@domwoe Can you point me at where you're running into an issue there?

libindy_vdr/src/vdr/vdr.rs
```diff
@@ -0,0 +1,570 @@
...
+
+                                    // let result = ResolutionResult {
+                                    //     did_resolution_metadata: None,
+                                    //     did_document: diddoc,
+                                    //     did_document_metadata: md,
+                                    // };
+                                    // if !doc.diddoc_content.is_none() {
+                                    //     callback(Ok(serde_json::to_string_pretty(&result).unwrap()))
+                                    // }
+                                }
```
domwoe, line 498: This would be an example.

I do `GET_NYM` Tx and if there's no `diddoc_content` present I need to do another `GET_ATTRIB` tx to see if there's a legacy endpoint attached to this nym. However, I can't get the lifetime requirements for the callbacks to match.

andrewwhitehead:
At the moment the nodejs wrapper is incomplete, and is not part of the automated tests or packaging. The Go wrapper is also not published, although it is tested. Both need additional maintainers and I'm fine with not supporting the multi-ledger resolver there just yet.

libindy_vdr/src/vdr/vdr.rs
andrewwhitehead: Hm. The callback pattern shouldn't be necessary outside of the `ffi` module, I think you can just write it as an async function and avoid the boxing.

andrewwhitehead: Ah never mind, that comes from using a PoolRunner rather than a Pool.

c2bo:
The Go wrapper is usable as a library by existing in this repository and a go.mod file being present. It is used e.g. in https://github.com/hyperledger/aries-framework-go-ext/tree/main/component/vdr/indy. I can help adding the new/modified calls to the go wrapper later on if there is interest in using these features.

If we want to make the wrappers more user friendly (by providing features like network/pool management), wouldn't it make sense to create a new rust library for this on top of or next to indy-vdr? Otherwise you would need reimplementation of the same logic in the different wrappers. What about having something like `libindy_networks` that adds functionality like initializing networks from a github repository and handling pool refreshes? This library could then be used in the different wrappers but also natively in Rust.

PR #87, opened 2022-03-24
Feature/python wrapper with multiple ledger support

domwoe:
Added python wrapper for multi-ledger did resolution

Exposes three simple ways to set up a resolver:

1. Provide a list of namespaces, genesis files will be retrieved from did indy networks repo
2. Local did indy networks compatible folder structure
3. Autopilot: Library will try to fetch genesis files from did indy networks repo when a DID with an unknown namespace is encountered.

### DIFF
--- a/wrappers/python/indy_vdr/resolver.py
+++ b/wrappers/python/indy_vdr/resolver.py
@@ -14,7 +14,7 @@ DID_PATTERN = re.compile(f"^{DID_INDY_PREFIX}:{NAMESPACE_PATTERN}:(.*)")
 
 
 class Resolver:
-    """did:indy compliant resolver interface
+    """did:indy compliant resolver interface.
 
     Args:
         pool_map: Dict mapping ledger namespaces to pool instances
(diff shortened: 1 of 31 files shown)


wrappers/python/indy_vdr/resolver.py
```diff
@@ -1,19 +1,132 @@
...
+                        
+                        else:
+
+                            services.append({
+                            "id": f"did:indy:{namespace}:{unqualified_did}#{service_type}",
+                            "type": service_type,
+                            "serviceEndpoint": service_endpoint 
+
+                            })
+                    result["didDocument"]["services"] = services;
```
domwoe, line 100: After countless hours trying to handle fetching the legacy endpoint (ATTRIB tx) on the rust side before crossing the ffi bridge I resigned.

I don't like it but I have added this code to the python wrapper to fetch the ATTRIB data and attach it to the did document.

If the `PoolResolver` is used from Rust projects (e.g. the proxy server) then this custom code is not necessary, but handled internally.

andrewwhitehead (approved):

PR #90, opened 2022-04-12
fix readme for indy-vdr-proxy

domwoe:
Fix of the doc to start indy vdr proxy with multiple ledgers flag.

Signed-off-by: Dominic Wörner <dom.woe@gmail.com>


dbluhm (approved):
Looks good to me

PR #88, opened 2022-04-11
feat(js): shared library

blu3beri:
- Shared interface library between the NodeJS side and the React Native side. Some minor TODO's are still included and will be resolved before merging.
- Does not include actual functionality, those will come in future PRs.
- (this is not going to main, but `javascript-wrapper`.

Signed-off-by: Berend Sliedrecht <berend@animo.id>

PR #91, opened 2022-04-25
feat(js): React Native

blu3beri:
Builds on top of #88 

- React Native wrapper implementation
- Uses TurboModules for improved performance and sync behaviour
- Has not been used in production yet, but it will be integrated and tested in AFJ

@TimoGlastra Do you think it is smart if you review the TypeScript side?


wrappers/javascript/react-native/package.json
```diff
@@ -0,0 +1,81 @@
...
+    "format": "yarn prettier --write",
+    "prettier": "prettier --ignore-path .gitignore '**/*+(.js|json|ts|md|yml|yaml)'",
+    "build": "yarn clean && yarn compile",
+    "clean": "rimraf -rf ./lib",
+    "compile": "tsc -p tsconfig.build.json",
+    "example": "yarn --cwd example",
+    "release": "release-it"
+  },
+  "dependencies": {
+    "indy-vdr-shared": "../shared/indy-vdr-shared-v103.0.0.tgz"
```
TimoGlastra, line 66: this should not be a tar right?

wrappers/javascript/react-native/tsconfig.build.json
```diff
@@ -0,0 +1,10 @@
+{
+  "extends": "../tsconfig.build.json",
+  "compilerOptions": {
+    "outDir": "./build",
+    "skipLibCheck": true,
+    "types": ["react-native"]
+  },
+  "include": ["src/**/*"],
+  "exclude": ["example"]
```
TimoGlastra, line 9: I think this is not needed anymore right as the example will be at the same level of the react-native / node packages right?

wrappers/javascript/shared/yarn.lock
```diff
@@ -0,0 +1,449 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
TimoGlastra, line 1: I think we only need one yarn.lock in the root `javascript` directory

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
TimoGlastra, line 1: I think this can be removed

TimoGlastra (approved):
Few small nits, but overall LGTM!

andrewwhitehead:
I think we can close the other two PRs as this supersedes them?

Why is the name 'react-native-indy-vdr-react-native'?

blu3beri:
> I think we can close the other two PRs as this supersedes them?

Yes, we can close the other pull requests. 


> Why is the name 'react-native-indy-vdr-react-native'?

That happened during the setup of the project. I will rename any reference to 'indy-vdr-react-native'.

PR #104, opened 2022-10-25
fix verkey expansion

c2bo:
Fixes a bug that was taken over from the initial resolver prototype, see https://github.com/IDunion/indy-did-resolver/issues/26 for more details.

### DIFF
--- a/libindy_vdr/src/resolver/did_document.rs
+++ b/libindy_vdr/src/resolver/did_document.rs
@@ -1,5 +1,6 @@
 use crate::common::error::prelude::*;
 use crate::ledger::responses::Endpoint;
+use indy_utils::base58;
 use serde::{Deserialize, Serialize};
 use serde_json::{self, Value as SJsonValue};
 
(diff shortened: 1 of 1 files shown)


c2bo:
failing step in GHA should be fixed by #105

libindy_vdr/src/resolver/did_document.rs
```diff
@@ -173,12 +174,52 @@ fn merge_diddoc(base: &mut SJsonValue, content: &SJsonValue) {
...
-fn expand_verkey(id: &str, verkey: &str) -> String {
-    if let Some(stripped) = verkey.strip_prefix('~') {
-        format!("{}{}", id, stripped)
+// Returns raw verkey in case of conversion errors, otherwise 'default' indy handling
+pub fn expand_verkey(id: &str, verkey: &str) -> String {
+    let expanded_verkey = expand_verkey_indy(id, verkey);
+    if expanded_verkey.is_err() {
+        return verkey.to_string();
+    }
+    return expanded_verkey.unwrap();
```
blu3beri, line 183: I think a match would be cleaner here:

```suggestion
    match expand_verkey_indy(id, verkey) {
      Ok(key) => key,
      Err(_) => verkey.to_string()
    }
```

returning with a `&str` would make it even terser though.

libindy_vdr/src/resolver/did_document.rs
```diff
@@ -173,12 +174,52 @@ fn merge_diddoc(base: &mut SJsonValue, content: &SJsonValue) {
...
+            0 | 1 => {
+                return Err(input_err("Unexpected verkey format"));
+            }
+            _ => {
+                let key = vec[0];
+                let key_type = vec[vec.len() - 1];
+                if key.len() == 0 || key_type.len() == 0 {
+                    return Err(input_err("Unexpected verkey format"));
+                }
+                (key, Some(key_type))
```
blu3beri, line 200:
```suggestion
  match (vec.first(), vec.last()) {
      (key, key_type) => Ok((key, key_type)),
      _ => Err(input_err("Unexpected verkey format"))
  }
```

Might require some additional code changes as it returns a `Ok(_)` now.

libindy_vdr/src/resolver/did_document.rs
```diff
@@ -173,12 +174,52 @@ fn merge_diddoc(base: &mut SJsonValue, content: &SJsonValue) {
...
+    }
+    return expanded_verkey.unwrap();
+}
+
+pub fn expand_verkey_indy(id: &str, verkey: &str) -> VdrResult<String> {
+    // separate optional crypto_type from verkey part
+    let (key, key_type) = if verkey.contains(":") {
+        let vec: Vec<&str> = verkey.split(":").collect();
+        match vec.len() {
+            0 | 1 => {
```
blu3beri, line 191: You check here for a length of 0 or 1, but you never check the total element count. In the match case below `vec[vec.len() - 1]` is called which might be the 99th element. I think a length check here would be a good addition.

blu3beri (changes requested):
I am not too familiar with the specific implementation so my review is mainly about the code.

libindy_vdr/src/resolver/did_document.rs
c2bo: Thanks! I fully agree - looking at your code feels like looking at rust code - mine feels like Go :D

libindy_vdr/src/resolver/did_document.rs
c2bo: I will change that whole key_type part to expect ecactly 2 outputs and error on anything else - the indy-sdk implementation kinda expects the same.

libindy_vdr/src/resolver/did_document.rs
c2bo: From a short test, this change would not catch empty strings and create different behavior

blu3beri: Yes you are correct

libindy_vdr/src/resolver/did_document.rs
blu3beri:
```rust
expand_verkey_indy(id, verkey).unwrap_or(verkey.to_string)
```

Would probably work as well now that I think about it.

blu3beri (approved):

andrewwhitehead (approved):
There are still a few odd code patterns for Rust but I think the behaviour is correct, and it can be cleaned up a bit later.

PR #110, opened 2022-11-07
Add support for config flag transactions

c2bo:
Implements new transaction added in https://github.com/hyperledger/indy-node/pull/1783 to indy-node 1.13.


c2bo:
This needs a test setup with indy-node 1.13 for functioning tests -> waiting for #112

PR #114, opened 2022-11-16
build(javascript): add javascript release pipeline

TimoGlastra:
This PR adds a release pipeline for the javascript wrapper for Indy VDR. Currently it is only set up to release the nodejs wrapper as we're still working on the iOS and Android builds. Will make a PR for those once ready.

To not publish all binaries within the NPM packages I've modified the release pipeline to publish the binaries to the Github release. I think this is useful beyond the JavaScript packages. When `indy-vdr-nodejs` is installed it will fetch the binaries from github and you're ready to use it.

When a release is made it will automatically add the binaries and release the JS / Python wrappers. there's also the option to publish the wrappers without publishing the binaries through the workflow dispatch. 

I've looked for a way to build the binaries for the v0.3.4 tag but seems I'm getting some errors. Do you think it would be possible to create a v0.3.5 release after this PR to bootstrap the publishing of the binaries?

What needs to happen form here:
- Before merge:
  - Add the `NPM_TOKEN` secret to this repository (@ryjones could you help out with this?) 
- After merge 
	- Create new tag and release

If you'd rather not create a new release I can look at manually building the old binaries at the time 0.3.4 was released and add those to the release.


TimoGlastra:
Can't request reviews.

@andrewwhitehead @blu3beri can you give this a review? Thanks

blu3beri (commented):
Will do a full review tomorrow. Just spotted this and I did not want to forget.

ryjones:
@TimoGlastra done

.github/workflows/build.yml
```diff
@@ -188,6 +251,68 @@ jobs:
...
+          (github.event_name == 'workflow_dispatch' && github.event.inputs.publish-wrappers == 'true')
+        run: |
+          echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" >> .npmrc
+          echo "registry=https://registry.npmjs.org/" >> .npmrc
+          echo "always-auth=true" >> .npmrc
+
+      # TODO: this will currently publish the version as defined in the package.json. If the version already
+      # exists on NPM it will skip the publishing. This means if a new version is released, but the version hasn't been
+      # updated in the package.json files yet, it won't publish to NPM (which is kinda nice). We should however add a check
+      # to see if the JS version matches the release version (do we want to keep the js and rust version in sync?)
```
blu3beri, line 309: We have discussed this externally as well, but my preference goes to no version sync. We will likely create many many patches when we integrate with AFJ and it would be I think rather annoying for the other packages. 

We can create a simple mapping in the README and we have to make sure that the latest indy-vdr is supported by the latest wrapper in JS.

wrappers/javascript/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+```sh
+yarn install
+```
+
+### Running Tests
+
+First make sure you've built the library using the `build.sh` script in the root of the repository.
+
+Then run the tests with the following command. Replace `<absolute-path>` with an absolute path to repository root.
```
blu3beri, line 24: relatives also work IIRC which makes the command static :). not exactly sure what the relative path is from where we want to run the tests. (maybe we can do `yarn run:with-build` that will add the env var to `yarn test`).

wrappers/javascript/nodejs/README.md
```diff
@@ -39,5 +30,7 @@ const getSchemaRequest = new GetSchemaRequest({
   schemaId: 'J6nTnUo3YLayzc2GUUctb1:2:MyName:1.0',
 })
 
-await pool.submitRequest({ requestHandle: getSchemaRequest.handle })
+const schemaResponse = await pool.submitRequest(getSchemaRequest)
```
blu3beri, line 33: We don't get type inference here, right? I think we should either add the solution we discussed previously or we should add the generic to `pool.sumbitRequest<T>` in the example.

wrappers/javascript/nodejs/package.json
```diff
@@ -47,10 +48,18 @@
     "typescript": "4.5.5"
   },
   "dependencies": {
-    "indy-vdr-shared": "*",
+    "@mapbox/node-pre-gyp": "^1.0.10",
     "ffi-napi": "^4.0.3",
+    "indy-vdr-shared": "0.1.0",
```
blu3beri, line 53: Always unsure about this, when using this locally, will this pick up the version inside your project or will it use the `0.1.0` from npm?

wrappers/javascript/shared/README.md
```diff
@@ -21,6 +23,37 @@ const getSchemaRequest = new GetSchemaRequest({
...
+
+```typescript
+// register.ts
+import 'indy-vdr-react-native'
+```
+
+```typescript
+// register.native.ts
+import 'indy-vdr-react-native'
+```
```
blu3beri, line 43: This confuses me slightly as you import the same package. Is that the point or am I missing something?

wrappers/javascript/shared/src/indyVdr/IndyVdrRequest.ts
```diff
@@ -62,4 +64,8 @@ export class IndyVdrRequest {
   public free(): void {
     indyVdr.requestFree({ requestHandle: this.handle })
   }
+
+  public get __responseType__(): ResponseType | undefined {
```
blu3beri, line 68: Ignore my comment about the generic in the docs, I see you have added it. Very nice!!

wrappers/javascript/shared/src/indyVdr/IndyVdrRequest.ts
```diff
@@ -62,4 +64,8 @@ export class IndyVdrRequest {
...
+    return undefined
+  }
```
blu3beri, line 70:
```suggestion
  public readonly __responseType__!: ResponseType
```

Might be a bit cleaner as it also omits the `undefined` which would also remove the `NonNullable` the from above. I vaguely remember that this worked, might have missed something. If the suggestion does not work, I am 100% fine with your solution.

blu3beri (changes requested):
Left some very small points. Quite some nice additions next to the release as well :).

wrappers/javascript/shared/src/indyVdr/IndyVdrRequest.ts
TimoGlastra: Event though it is an internal type, it can still be accessed publicly. So reason making it | undefined is that this way the type is not lying as the value could be undefined. By setting that it will always be defined it can lead to runtime errors.

wrappers/javascript/shared/README.md
TimoGlastra: Good catch. One should be nodejs :)

wrappers/javascript/nodejs/package.json
TimoGlastra: Yes this will work locally if the version matches the version of the package in the local workspace. Need to do this as otherwise yarn won't correctly increase the version for other packages.

wrappers/javascript/README.md
TimoGlastra: Yes, had some issues with this as the base path is kinda weird with yarn workspaces, but the custom command is smart!

.github/workflows/build.yml
TimoGlastra: Agreed, will update the command.

TimoGlastra:
Thanks @ryjones!

Resolved your feedback @blu3beri 

Over to you now @andrewwhitehead

wrappers/javascript/shared/src/types/IndyVdr.ts
```diff
@@ -84,7 +79,7 @@ export interface IndyVdr {
 
   buildGetTxnRequest(options: GetTransactionRequestOptions): number
 
-  buildGetValidatorInfoRequest(options: GetValidatorInfoRequestOptions): number
+  buildGetValidatorInfoRequest(options: GetValidatorInfoActionOptions): number
```
blu3beri, line 82: Why did you rename it to actions?

blu3beri (approved):

wrappers/javascript/shared/src/types/IndyVdr.ts
TimoGlastra: Because the getValidatorInfo should be run as action (If I understand the Python wrapper correctly):

https://github.com/hyperledger/indy-vdr/blob/d40a818dbee44e47476bc36a6f1a2249530a732f/wrappers/python/indy_vdr/pool.py#L70-L75

andrewwhitehead (dismissed):

andrewwhitehead:
@TimoGlastra The library version is already set to 0.4.0 on main, so we could do a release of that version.

TimoGlastra:
> The library version is already set to 0.4.0 on main, so we could do a release of that version. 

That would be great! I've updated the afj package to look for the binaries in the 0.4.0 release tag. If we merge this PR and create a release everything should be released and work as expected. 

This is now ready to be merged

andrewwhitehead (approved):

PR #116, opened 2022-12-01
updated ZMQ to the latest version

blu3beri:
- No more zmq vendored feature
- makes cross-compilation a lot easier as `libzmq3-dev` is not required anymore (this is required for the upcoming Android pipeline)
- Tests still pass 
- [changelog](https://github.com/erickt/rust-zmq/releases/tag/v0.10.0)

### DIFF
--- a/libindy_vdr/Cargo.toml
+++ b/libindy_vdr/Cargo.toml
@@ -18,10 +18,9 @@ crate-type = ["staticlib", "rlib", "cdylib"]
 [features]
 ffi = ["ffi-support", "logger"]
 logger = ["env_logger", "log"]
-zmq_vendored = ["zmq/vendored"]
 local_nodes_pool = []
 rich_schema = ["indy-data-types/rich_schema"]
-default = ["ffi", "log", "zmq_vendored"]
+default = ["ffi", "log"]
 
 [dependencies]
 env_logger = { version = "0.9", optional = true }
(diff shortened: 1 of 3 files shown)


blu3beri:
Hmm, looks like there is an issue with the Go wrapper. Will take a look tomorrow.

libindy_vdr/Cargo.toml
```diff
@@ -44,7 +43,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 sha3 = "0.10"
 thiserror = "1.0"
-zmq = "0.9"
+zmq = { package = "zmq2", git = "https://github.com/erickt/rust-zmq.git", rev = "5821ee121c30311d961e18b9ff7017603c600548" }
```
andrewwhitehead, line 46: I think the zmq 0.10.0 release from last month supercedes zmq2?

blu3beri: Yes, I was trying to fix an issue with the go wrapper (and likely all the other wrappers). We call a function from the zmq library that was removed when updating to v0.10.0. Issue: https://github.com/erickt/rust-zmq/issues/357#issue-1474129334

However this fix breaks something on the windows build. 

If [this issue ](https://github.com/erickt/rust-zmq/issues/357#issue-1474129334) is resolved it might be good to move towards 0.10.0, but right now I will try to do some cross-compilation work to get it working with v0.9.0.

PR #118, opened 2022-12-06
build(android): add CI/CD

blu3beri:
- Will build the android wrapper using cross
  - The default image used by cross has some issues so it has to be specified in the `Cross.toml` file that we use the `main` tag for every image.
- Will create an artifact called `android-jniLibs` (open for better naming), which is a combination of the architectures that are being created. Android uses this structure to have "one" package for every architecture.

Workflow has been tested, without the `publish-release` input. Result can be found here: https://github.com/blu3beri/indy-vdr/actions/runs/3628442119

Work funded by the Government of Ontario.

### DIFF
--- a/wrappers/javascript/shared/src/indyVdr/IndyVdrRequest.ts
+++ b/wrappers/javascript/shared/src/indyVdr/IndyVdrRequest.ts
@@ -17,10 +17,14 @@ export type RequestSetEndorserOptions = {
   endorser: string
 }
 
-export type RequestResponseType<Request extends IndyVdrRequest> = NonNullable<Request['__responseType__']>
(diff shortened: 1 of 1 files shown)


TimoGlastra (approved):
LGTM. I have a branch somewhere that combines the android and ios libs into a custom directory that we can use with node-pre-gyp. Will find that for the ios build pr

TimoGlastra (approved):
Wondering how we will add it to the release if it is in a seperate pipeline? We can maybe just uplaod them to to release separately...

blu3beri:
> Wondering how we will add it to the release if it is in a seperate pipeline? We can maybe just uplaod them to to release separately...

Yeah, might have to move it to the `build.yml` file. Would be nice if we have some separation between the different flow, but depending on each other would be a bit more annoying.

PR #119, opened 2022-12-06
Update build actions

andrewwhitehead:
- Update to node16-compatible versions of actions
- Use cargo-cross to build linux binaries
- Add linux/arm64 build
- Use Python 3.8 over 3.6 for testing (3.6 is no longer easily available)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,16 +30,13 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Install Rust toolchain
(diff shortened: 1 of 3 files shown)


blu3beri (approved):

blu3beri:
also, is there a specific reason for using dtolnay's toolchain action? Never used it myself, are there any benefits?

andrewwhitehead:
The actions-rs toolchain action isn't really maintained, and it raises errors due to being on node 12. The dtolnay action normally specifies the Rust toolchain version in the `@branch` tag, but when it's a variable we set that to `master` and pass in the toolchain parameter instead.

blu3beri (approved):

PR #120, opened 2022-12-12
ci: use ubuntu 20

TimoGlastra:
Use ubuntu 20 in ci to fix issue of python 3.6 not being found.

Relevant issue: https://github.com/actions/setup-python/issues/162


TimoGlastra:
Ah I see #119 does this and more...

PR #121, opened 2022-12-12
refactor(javascript): better response type inference

TimoGlastra:
Small improvement to type response type inference for requests. Using the `infer` key we don't have to make the __responseType__ property public.

### DIFF
--- a/wrappers/javascript/shared/src/indyVdr/IndyVdrRequest.ts
+++ b/wrappers/javascript/shared/src/indyVdr/IndyVdrRequest.ts
@@ -17,10 +17,14 @@ export type RequestSetEndorserOptions = {
   endorser: string
 }
 
-export type RequestResponseType<Request extends IndyVdrRequest> = NonNullable<Request['__responseType__']>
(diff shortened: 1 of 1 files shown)


wrappers/javascript/shared/src/indyVdr/IndyVdrRequest.ts
```diff
@@ -17,11 +17,15 @@ export type RequestSetEndorserOptions = {
   endorser: string
 }
 
-export type RequestResponseType<Request extends IndyVdrRequest> = NonNullable<Request['__responseType__']>
+export type RequestResponseType<Request> = Request extends IndyVdrRequest<infer ResponseType> ? ResponseType : never
```
blu3beri, line 20: Shouldn't it still be nonnullable?

blu3beri (approved):

wrappers/javascript/shared/src/indyVdr/IndyVdrRequest.ts
TimoGlastra: No because we don't use a property anymore to infer the type, we infer it directly from the generic (using the `infer ResponseType`)

PR #126, opened 2023-01-04
integrate indy utils

berendsliedrecht:
- Integrated txn_signature into indy-vdr
- Added base64 and base58 inside the package
- Removed features from indy-utils

## Question

Do we want to just implement all the functionality, from `indy-utils` inside `indy-vdr` and reimplement this in `anoncreds` (There should not be a lot of code reimplemented inside anoncreds as most of it, if not all, will be stripped out)?


andrewwhitehead:
Looks good, we can probably also drop the hash feature and use sha2 directly (although it may have been done this way to avoid conflicts with Ursa's out of date version selection?). indy-utils might be replaced by a crate that only does Indy-style DID/key handling.

blu3beri:
> indy-utils might be replaced by a crate that only does Indy-style DID/key handling.

Yeah that is an interesting one. I don’t think we need any reusability within the packages, or is it required in askar as well? I think if we do not reuse it anywhere it might be fine to have it inside `indy-vdr`, but I can see a benefit in a seperate crate (just wondering a bit about the location… could be `indy-shared-rs`.

andrewwhitehead:
I think instead of moving indy-utils we can just copy over the remaining bits that are used. I believe we can do without the ed25519 key generation, it seems to be only used in tests, and so there's just the DidValue and ShortDidValue structs that are used in various places here.

For indy-data-types, it probably makes sense to copy those over as well, as was done in anoncreds-rs. We don't need the Ursa integration in this case. The biggest issue would be for projects that are using both crates (indy-vdr and either indy-credx or anoncreds-rs) and trying to create a data object then submit it to the ledger. With the existing code you could create a new CredentialDefinition and pass it directly to indy-vdr for publishing, but after this change it would need to be serialized and deserialized into the new type. I'm okay with that, because the ledger representation is not guaranteed to match the internal format going forward.

andrewwhitehead:
I think the only part of Ursa that's being used at the moment is the BLS signature support, so it might make sense to import that as well to avoid complications. The implementation is currently using unmaintained dependencies and it's on a curve with fairly weak security (although we can't change that part right now).

andrewwhitehead:
Added these changes into the increasingly large #166

PR #132, opened 2023-01-12
Update deprecated actions

andrewwhitehead:


swcurran:
Does this PR fix the failures in the Builds that happened when running the existing GHAs, or are the updates in this PR to the build actions failing?   I would assume that it is the latter, and those need to be fixed before we can build.

andrewwhitehead:
@swcurran I don't think the current builds are failing, although it did fail to publish the javascript packages due to a missing API token.

blu3beri (approved):

PR #133, opened 2023-01-13
build(javascript): fix some small release issues

TimoGlastra:
There were some small issues with the JavaScript release. The publishing of the binaries fully worked 🎉

There's one issue before we can publish again and that is that the Indy VDR Shared package is already taken on NPM: https://www.npmjs.com/package/indy-vdr-shared. It's in security hold. I've submitted a request to Github/NPM to release the package and give us ownership. That can take weeks. I suggest we don't wait weeks, but maybe until next week. Otherwise we can use `@hyperledger` prefix for the package names.

### DIFF
--- a/wrappers/javascript/indy-vdr-nodejs/package.json
+++ b/wrappers/javascript/indy-vdr-nodejs/package.json
@@ -53,7 +53,7 @@
   "binary": {
     "module_name": "indy_vdr",
     "module_path": "native",
-    "remote_path": "v0.4.0",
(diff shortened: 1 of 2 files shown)


blu3beri (approved):
I am okay with using `@hyperledger/` as the prefix for the packages. They might get a little bit long, but that's only a small issue with installing. Lets wait a week as you said and otherwise just go to `@hyperledger/indy-vdr-*`

andrewwhitehead (approved):

PR #135, opened 2023-01-21
fix(js): typing issues

TimoGlastra:
Fixes some minor typing issues

PR #136, opened 2023-01-21
chore(js): add @hyperledger prefix

TimoGlastra:
Adds the `@hyperledger` prefix to all js packages and updates the version to `0.1.0-dev.2`. 

After this PR and #135 are merged would like to make another release.

I think we can just run the ci pipeline on main, do not publish binaries, but do public wrappers. Because python version insn't incresed I think it will just skip them?

### DIFF
--- a/wrappers/javascript/indy-vdr-nodejs/README.md
+++ b/wrappers/javascript/indy-vdr-nodejs/README.md
@@ -15,7 +15,7 @@ yarn add @hyperledger/indy-vdr-nodejs
 
 ## Usage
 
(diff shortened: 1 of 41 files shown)


blu3beri (approved):

andrewwhitehead:
Might as well update the python wrapper version as well I think, to keep them consistent?

TimoGlastra:
> Might as well update the python wrapper version as well I think, to keep them consistent?

Done

blu3beri (approved):

PR #134, opened 2023-01-19
Added rust methods to build FREEZE_LEDGER and GET_FROZEN_LEDGER_REQUESTS

Artemkaaas:
Updated `Request Builder` to provide methods for building `FREEZE_LEDGER` and `GET_FROZEN_LEDGER_REQUESTS` requests. 
Currently, methods are only available in RUST API because it seems that Ledger Admin specific methods are not exposed in C / wrappers API.

PR #137, opened 2023-01-23
Added missing reqeust builders into FFI and Python wrapper

Artemkaaas:
Here is the list of added functions:
```
build_node_request
build_pool_config
build_pool_restart
build_auth_rule_request
build_auth_rules_request
build_get_auth_rule_request
build_ledgers_freeze_request
build_get_frozen_ledgers_request 
```

These methods are needed for upgrading the Indy Test Automation repository to use Indy VDR for testing Indy Plenum and Indy Node instead of libindy.

This PR includes and depends on the changes from the PR https://github.com/hyperledger/indy-vdr/pull/134

### DIFF
--- a/libindy_vdr/src/ffi/ledger.rs
+++ b/libindy_vdr/src/ffi/ledger.rs
@@ -630,10 +630,10 @@ pub extern "C" fn indy_vdr_build_node_request(
 }
 
 #[no_mangle]
-pub extern "C" fn indy_vdr_build_pool_config(
+pub extern "C" fn indy_vdr_build_pool_config_request(
     identifier: FfiStr,
-    writes: bool,
-    force: bool,
+    writes: i8,
+    force: i8,
     handle_p: *mut RequestHandle,
 ) -> ErrorCode {
     catch_err! {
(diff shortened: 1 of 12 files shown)


libindy_vdr/src/ffi/ledger.rs
```diff
@@ -602,3 +604,196 @@ pub extern "C" fn indy_vdr_build_get_rich_schema_object_by_metadata_request(
...
+        let handle = add_request(req)?;
+        unsafe {
+            *handle_p = handle;
+        }
+        Ok(ErrorCode::Success)
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn indy_vdr_build_pool_config(
```
andrewwhitehead, line 633:
```suggestion
pub extern "C" fn indy_vdr_build_pool_config_request(
```

libindy_vdr/src/ffi/ledger.rs
```diff
@@ -602,3 +604,196 @@ pub extern "C" fn indy_vdr_build_get_rich_schema_object_by_metadata_request(
...
+    identifier: FfiStr,
+    writes: bool,
```
andrewwhitehead, line 635: Can we use i8/c_int8 here? I'm not sure we can rely on both Rust's bool and c_bool being the same size.

andrewwhitehead:
I think this supersedes #134?

andrewwhitehead (approved):

PR #145, opened 2023-01-26
fix(js): framework path in podspec file

TimoGlastra:
So close!


blu3beri:
Ah good catch! 

However I think we should change the lib prefix on the xcframework. 

Here (https://github.com/hyperledger/indy-vdr/blob/4e2d1c1cf81f0e3e9e2b1607658d98ec47daee66/.github/workflows/build.yml#L291) it should be `indy_vdr.xcframework`.

A bit more conventional for xcframeworks.

blu3beri (approved):

PR #147, opened 2023-01-27
fix(js): request type inference

TimoGlastra:
Signed-off-by: Timo Glastra <timo@animo.id>


wrappers/javascript/indy-vdr-shared/src/indyVdr/IndyVdrRequest.ts
```diff
@@ -23,8 +23,8 @@ export class IndyVdrRequest<ResponseType extends Record<string, unknown> = Recor
   private _handle: number
 
   // We need to use the generic that is passed to this class, otherwise TypeScript will lose the generic type passed to IndyVdrRequest
-  // and we cant' infer the response type. The value is private, so it's won't be accessible from outside the class.
-  private __responseType__?: ResponseType
+  // and we cant' infer the response type. The value is protected, so it's won't be accessible from outside the class.
```
blu3beri, line 26:
```suggestion
  // and we can't infer the response type. The value is protected, so it's won't be accessible from outside the class.
```

blu3beri (approved):

blu3beri (approved):

PR #112, opened 2022-11-10
Reuse indy-node-container image

c2bo:
Implements #111

PR #166, opened 2023-03-01
Merge did:indy support

andrewwhitehead:
Currently, tests are failing with the JS wrapper due to a timeout in GetAttribRequest. I think this might be a compatibility issue with the did:indy-supporting indy-node version, but I'm not certain.


swcurran:
Who can/should look at this?  @blu3beri — is that something you can look at?

@andrewwhitehead — any details that would help — any details on how to reproduce, what is happening/where?

andrewwhitehead:
There are more details in the CI logs, specifically

```
FAIL tests/GetAttribRequest.test.ts (120.218 s)
  GetAttribRequest
    ✕ Submit request (120001 ms)

  ● GetAttribRequest › Submit request

    thrown: "Exceeded timeout of 120000 ms for a test.
    Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."

      10 |   beforeAll(() => (pool = setupPool()))
      11 |
    > 12 |   test('Submit request', async () => {
         |   ^
      13 |     const request = new GetAttribRequest({ targetDid: DID, raw: 'endpoint' })
      14 |     const response: GetAttribResponse = await pool.submitRequest(request)
      15 |

      at tests/GetAttribRequest.test.ts:12:3
      at Object.<anonymous> (tests/GetAttribRequest.test.ts:7:1)
```

But there also seems to be a segfault after the tests complete (could be related to the Python test shutdown issue where it needs to stop the tokio threads first, or could be the additional parameters to the `indy_vdr_build_nym_request` FFI method).

c2bo:
This still has a custom old build of indy-node in it. (from here: https://github.com/domwoe/indy-node/tree/feature/did-indy-new). Should we test #112 on this branch and figure out of this fixes things?

c2bo:
Hmm, I tried testing this locally with the current and updated docker image and it seems to have no impact (both times the GetAttribRequest timeouts). Interesting is that I see another test failing:
```
 FAIL  tests/GetNymRequest.test.ts
  ● GetNymRequest › Submit request

    Invalid pointer for result value

      14 |   const indyVdrErrorObject = JSON.parse(nativeError.deref() as string) as IndyVdrErrorObject
      15 |
    > 16 |   throw new IndyVdrError(indyVdrErrorObject)
         |         ^
      17 | }
      18 |

      at handleError (src/error.ts:16:9)
      at NodeJSIndyVdr.buildGetNymRequest (src/NodeJSIndyVdr.ts:267:5)
      at new buildGetNymRequest (../indy-vdr-shared/src/builder/GetNymRequest.ts:28:28)
      at _callee$ (tests/GetNymRequest.test.ts:13:21)
      at tryCatch (../node_modules/regenerator-runtime/runtime.js:63:40)
      at Generator.invoke [as _invoke] (../node_modules/regenerator-runtime/runtime.js:294:22)
      at Generator.next (../node_modules/regenerator-runtime/runtime.js:119:21)
      at tryCatch (../node_modules/regenerator-runtime/runtime.js:63:40)
      at invoke (../node_modules/regenerator-runtime/runtime.js:155:20)
      at ../node_modules/regenerator-runtime/runtime.js:190:11
```

c2bo:
Nevermind - the typescript tests are running with the sovrin buildernet instead of the local pool setup? 
The libindy_vdr header (in libindy_vdr/include) file seems to be the old one before the changes - shouldn't this always create problems with a changed interface?

andrewwhitehead:
@c2bo Not sure if the separate did-indy-pool.dockerfile is still needed or we can just use indy-node-container 1.13.2-rc5 for all tests?

c2bo:
> @c2bo Not sure if the separate did-indy-pool.dockerfile is still needed or we can just use indy-node-container 1.13.2-rc5 for all tests?

We don't need the separate dockerifle anymore as far as i know -> That would be fixed by #112. I will re-base to this branch and create a new PR

TimoGlastra:
We'll try to update the js wrapper to use the local pool instead!

c2bo:
I added a small bash script in #169 that we can use for the different tests that want to use a local node pool. I just realized the same is true for the python wrapper - that one currently uses the idunion test network for its tests
```
./test.sh up
./test.sh down
```

test.sh
```diff
@@ -0,0 +1,50 @@
+#/bin/bash
```
berendsliedrecht, line 1: Do we want to keep this file? If so, might be good to move it inside a scripts directory.

libindy_vdr/src/common/error.rs
```diff
@@ -46,6 +46,8 @@ pub enum VdrErrorKind {
     PoolRequestFailed(String),
     #[error("Pool timeout")]
     PoolTimeout,
+    #[error("Resolver error ")]
```
berendsliedrecht, line 49:
```suggestion
    #[error("Resolver error")]
```

libindy_vdr/src/ffi/resolver.rs
```diff
@@ -0,0 +1,79 @@
...
+#[no_mangle]
+pub extern "C" fn indy_vdr_resolve(
+    pool_handle: PoolHandle,
+    did: FfiStr,
+    cb: Option<extern "C" fn(cb_id: i64, err: ErrorCode, response: *const c_char)>,
+    cb_id: i64,
+) -> ErrorCode {
+    catch_err! {
+    trace!("Resolve DID: {:#?}", did);
+    let cb = cb.ok_or_else(|| input_err("No callback provided")).unwrap();
```
berendsliedrecht, line 18:
```suggestion
    let cb = cb.ok_or_else(|| input_err("No callback provided"))?;
```

libindy_vdr/src/ffi/resolver.rs
```diff
@@ -0,0 +1,79 @@
...
+                Ok(result) => (ErrorCode::Success, result),
+                Err(err) => {
+
+                    let code = ErrorCode::from(err.kind());
+                    set_last_error(Some(err));
+                    (code, String::new())
+
+                }
+            };
+            cb(cb_id, errcode, rust_string_to_c(reply))
```
berendsliedrecht, line 38:
```suggestion
        Box::new(move |ledger_reply| {
            let (errcode, reply) = match handle_resolution_result(ledger_reply, did) {
                Ok(result) => (ErrorCode::Success, result),
                Err(err) => {
                    let code = ErrorCode::from(err.kind());
                    set_last_error(Some(err));
                    (code, String::new())
                }
            };
            cb(cb_id, errcode, rust_string_to_c(reply))
```

libindy_vdr/src/ffi/resolver.rs
```diff
@@ -0,0 +1,79 @@
...
+#[no_mangle]
+pub extern "C" fn indy_vdr_dereference(
+    pool_handle: PoolHandle,
+    did_url: FfiStr,
+    cb: Option<extern "C" fn(cb_id: i64, err: ErrorCode, response: *const c_char)>,
+    cb_id: i64,
+) -> ErrorCode {
+    catch_err! {
+    trace!("Dereference DID Url: {:#?}", did_url);
+    let cb = cb.ok_or_else(|| input_err("No callback provided")).unwrap();
```
berendsliedrecht, line 54:
```suggestion
    let cb = cb.ok_or_else(|| input_err("No callback provided"))?;
```

libindy_vdr/src/ffi/resolver.rs
```diff
@@ -0,0 +1,79 @@
...
+
+                    let (errcode, reply) = match handle_resolution_result(ledger_reply, did_url) {
+                        Ok(result) => (ErrorCode::Success, result),
+                        Err(err) => {
+
+                            let code = ErrorCode::from(err.kind());
+                            set_last_error(Some(err));
+                            (code, String::new())
+
+                        }
```
berendsliedrecht, line 72:
```suggestion
                Box::new(move |ledger_reply| {
                    let (errcode, reply) = match handle_resolution_result(ledger_reply, did_url) {
                        Ok(result) => (ErrorCode::Success, result),
                        Err(err) => {
                            let code = ErrorCode::from(err.kind());
                            set_last_error(Some(err));
                            (code, String::new())
                        }
```

libindy_vdr/src/resolver/resolver.rs
```diff
@@ -0,0 +1,203 @@
...
+        } else {
+            None
+        };
+
+        let diddoc = match data {
+            Result::DidDocument(mut doc) => {
+                // Try to find legacy endpoint using a GET_ATTRIB txn if diddoc_content is none
+                if doc.diddoc_content.is_none() {
+                    let (seq_no, timestamp) = if let Some(md) = &md {
+                        let j_seq_no = &md.node_response["result"]["seqNo"];
```
berendsliedrecht, line 67: More out of curiousity, what is the `j_` prefix?

libindy_vdr/src/resolver/utils.rs
```diff
@@ -0,0 +1,429 @@
...
+                object_type: String::from("Unknown"),
+            }),
+        ),
+    })
+}
+
+pub fn parse_ledger_data(ledger_data: &str) -> VdrResult<(SJsonValue, SJsonValue, SJsonValue)> {
+    let v: SJsonValue = serde_json::from_str(ledger_data)
+        .map_err(|_| err_msg(VdrErrorKind::Resolver, "Could not parse ledger response"))?;
+    // Unwrap should be safe here
```
berendsliedrecht, line 230: Nothing is being unwrapped anymore.

libindy_vdr/tests/nym.rs
```diff
@@ -253,6 +272,240 @@ mod send {
...
+                .unwrap()
+                .as_str()
+                .unwrap(),
+        )
+        .unwrap();
+
+        assert_eq!(expected_data, serde_json::to_value(data).unwrap());
+    }
+
+    // FIXME: Uncomment as soon as indy-utils is updated
```
berendsliedrecht, line 452: indy-utils did receive an update. Can we uncomment this of do we have a to keep it commented?

libindy_vdr/tests/resolver.rs
```diff
@@ -0,0 +1,109 @@
...
+        let v: serde_json::Value = serde_json::from_str(result.as_str()).unwrap();
+
+        let diddoc = &v["didDocument"];
+        let metadata = &v["didDocumentMetadata"];
+
+        assert_eq!("NYM", metadata["objectType"]);
+        assert_ne!(&serde_json::Value::Null, diddoc)
+    }
+
+    // #[rstest]
```
berendsliedrecht, line 72: Can this be removed as well?

libindy_vdr/tests/utils/fixtures.rs
```diff
@@ -42,8 +45,42 @@ pub fn fq_my_did() -> DidValue {
 }
 
 #[fixture]
+// did:sov self-certified DID
 pub fn identity() -> Identity {
-    Identity::new(None)
+    Identity::new(None, None)
+}
+
+// FIXME: Uncomment as soon as indy-utils is updated
```
berendsliedrecht, line 53: Same comment here re `indy-utils`.

berendsliedrecht (approved):
Couple small items, nothing blocking.

libindy_vdr/src/resolver/resolver.rs
andrewwhitehead: I think it's short for json, since they are json nodes.

swcurran:
w00t!!!

PR #168, opened 2023-03-02
Build updates

andrewwhitehead:
- Update the base images used in `cross` builds
- Verify that Python wheels (and by extension npm modules) are compatible with manylinux 2014
- Only use build.sh (now build-universal.sh) for MacOS universal builds, otherwise prefer `cargo build`
- Improve caching of Rust dependencies
- Minor adjustments to build-xcframework script


blu3beri (dismissed):

andrewwhitehead:
The selected `time` and `env_logger` dependencies require at least Rust 1.63. `time` could be fixed to 0.3.17 for Rust 1.60 support (it's only used in tests) and `env_logger` could easily be downgraded.

PR #173, opened 2023-03-14
fix(js): defer library loading for nodejs

genaris:
Prevent native library from being loaded right when `@hyperledger/indy-vdr-nodejs` is imported by deferring its loading to when it's actually used. 

This is a problem we found during AFJ test suites, which import the library in general purpose test helpers and make the native library to be loaded more times than needed, allocating more memory and making the suite to crash.


wrappers/javascript/indy-vdr-nodejs/src/library/register.ts
```diff
@@ -72,4 +72,8 @@ const getLibrary = () => {
 
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore
```
blu3beri, line 74: Both of these can be removed now.

wrappers/javascript/indy-vdr-nodejs/src/error.ts
```diff
@@ -3,13 +3,13 @@ import type { IndyVdrErrorObject } from '@hyperledger/indy-vdr-shared'
 import { IndyVdrError } from '@hyperledger/indy-vdr-shared'
 
 import { allocateString } from './ffi'
-import { nativeIndyVdr } from './library'
+import { getNativeIndyVdr } from './library'
 
 export const handleError = (code: number) => {
```
blu3beri, line 8: Seeing this now, I would say it makes more sense to have `handleError` as a private method on the `NodeJSIndyVdr` class. Could you include that in the PR?

blu3beri (changes requested):
Two main comments, otherwise its a very minimal change that would be a good addition.

blu3beri (approved):
LGTM!

PR #176, opened 2023-03-17
fix(js): add ref-array-di types

TimoGlastra:


blu3beri (approved):
Do we know for sure this is going to work?

PR #200, opened 2023-08-09
Switch to indy-blssignatures

andrewwhitehead:
This removes Ursa as a dependency.

There are a few fixes for warnings included, and a new test that exercises the state proof validation.


libindy_vdr/src/pool/manager.rs
```diff
@@ -0,0 +1,212 @@
...
+
+/// The default `Pool` implementation
+#[derive(Clone)]
+pub struct PoolImpl<S: AsRef<PoolSetup> + Clone, T: Networker + Clone> {
+    setup: S,
+    networker: T,
+}
+
+/// A verifier pool instance restricted to a single thread
+pub type LocalPool = PoolImpl<Rc<PoolSetup>, LocalNetworker>;
```
berendsliedrecht, line 73: What is the reason for having a single and multi threaded pool? I assume the performance between the arc and rc is negligible here.

berendsliedrecht (approved):

libindy_vdr/src/pool/manager.rs
andrewwhitehead: I think it was mostly to allow for a single threaded web server, but I agree the performance wouldn't matter here. None of that code has changed though it's just that the file was renamed.

Issue #206, opened 2023-08-10
Mismatch arguments requirements for GetAcceptanceMechanismsRequestOptions in indy-vdr-shared and indy-vdr-react-native/cpp


mrlunin:
I was playing with issuing schemas, credential definition and credentials from react-native app in AriesFrameworkjavascript but always receive the error:
`Error: Value 'submitterDid' is not of type string\n    at buildGetAcceptanceMechanismsRequest (native)`

submitterDid is optional at GetAcceptanceMechanismsRequestOptions defined in indy-vdr-shared but this parameter is required in cpp code here:

GetAcceptanceMechanismsRequestOptions: https://github.com/hyperledger/indy-vdr/blob/main/wrappers/javascript/indy-vdr-shared/src/builder/GetAcceptanceMechanismsRequest.ts#L7

Cpp code: https://github.com/hyperledger/indy-vdr/blob/main/wrappers/javascript/indy-vdr-react-native/cpp/indyVdr.cpp#L67

at line 
 auto submitterDid = jsiToValue<std::string>(rt, options, "submitterDid");

PR #207, opened 2023-08-10
Mismatch arguments requirements for GetAcceptanceMechanismsRequestOptions in indy-vdr-shared and indy-vdr-react-native/cpp

mrlunin:
Ticket https://github.com/hyperledger/indy-vdr/issues/206


berendsliedrecht:
Thanks for the fix. Could you sign off your commit so the DCO passes? If that all checks it, I will merge it.

mrlunin:
@berendsliedrecht ok, done

andrewwhitehead (approved):

PR #211, opened 2023-08-14
Issue #210 InvalidClientTaaAcceptanceError time too precise error if container timezone is not UTC

Ennovate-com:
Adds timezone.utc to datetime.combine to ensure time will be midnight in UTC regardless of the default timezone

### DIFF
--- a/wrappers/python/indy_vdr/ledger.py
+++ b/wrappers/python/indy_vdr/ledger.py
@@ -761,7 +761,9 @@ def prepare_txn_author_agreement_acceptance(
     if not accepted_time:
         # rough timestamp
         accepted_time = int(
-            datetime.combine(date.today(), datetime.min.time(), tzinfo=timezone.utc).timestamp()
+            datetime.combine(
+                date.today(), datetime.min.time(), timezone.utc
+            ).timestamp()
         )
     result = lib_string()
     do_call(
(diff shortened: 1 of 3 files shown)


wrappers/python/indy_vdr/ledger.py
```diff
@@ -760,7 +761,9 @@ def prepare_txn_author_agreement_acceptance(
     if not accepted_time:
         # rough timestamp
         accepted_time = int(
-            datetime.combine(date.today(), datetime.min.time()).timestamp()
+            datetime.combine(
+                date.today(), datetime.min.time(), timezone.utc
```
andrewwhitehead, line 765: Could you change this to datetime.timezone.utc? This library doesn't import pytz.

Ennovate-com: @andrewwhitehead Yes, changed, Thanks! Let me know if there are any other changes needed.

wrappers/python/indy_vdr/ledger.py
```diff
@@ -760,7 +760,9 @@ def prepare_txn_author_agreement_acceptance(
     if not accepted_time:
         # rough timestamp
         accepted_time = int(
-            datetime.combine(date.today(), datetime.min.time()).timestamp()
+            datetime.combine(
+                date.today(), datetime.min.time(), datetime.timezone.utc
```
andrewwhitehead, line 764:
```suggestion
                date.today(), datetime.min.time(), timezone.utc
```

andrewwhitehead (approved):

PR #212, opened 2023-08-17
fix(js)!: use patched ref-napi/ffi-napi for nodejs

genaris:
Similar approach as https://github.com/hyperledger/anoncreds-rs/pull/231 to make an easier set up in Node.js environment.

### DIFF
--- a/wrappers/javascript/package.json
+++ b/wrappers/javascript/package.json
@@ -44,8 +44,5 @@
     "eslint-plugin-prettier": "^4.0.0",
     "prettier": "^2.5.1",
     "typescript": "~4.9.4"
-  },
-  "engines": {
-    "node": ">= 18"
   }
 }
--- a/wrappers/javascript/package.json
+++ b/wrappers/javascript/package.json
@@ -46,6 +46,6 @@
     "typescript": "~4.9.4"
   },
   "engines": {
-    "node": ">= 12"
+    "node": ">= 18"
   }
 }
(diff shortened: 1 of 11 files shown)


TimoGlastra (dismissed):

wrappers/javascript/package.json
```diff
@@ -46,6 +46,6 @@
     "typescript": "~4.9.4"
   },
   "engines": {
-    "node": ">= 12"
+    "node": ">= 18"
```
berendsliedrecht, line 49: I don't think that the root workspace package.json engine influences anything when a user installs `@hyperledger/indy-vdr-nodejs`. I think it should be done in the `indy-vdr-nodejs` package only as well. The shared and react native packages can be used with lower node versions.

genaris: Right, actually the restriction will come from the dependencies (patched `ref-napi` and `ffi-napi`) when attempting to do `yarn install`. Anyway, I'll move it to `indy-vdr-nodejs` to make it clear from there.

TimoGlastra (approved):

TimoGlastra:
Do we want to release a 0.2.x for indy-vdr that only supports nodejs 18 right away?

We should probably also add a version compatiblity table to the readme: https://github.com/hyperledger/aries-askar/tree/main/wrappers/javascript/aries-askar-nodejs#version-compatibility

andrewwhitehead (approved):

andrewwhitehead:
@TimoGlastra We need to get a 0.4 release out of the Python wrapper as well, at the moment there's still no non-dev arm64 build. Is there anything else missing?

PR #215, opened 2023-09-11
Update versions to 0.4.0

andrewwhitehead:


swcurran:
Hmm…failing test.  @berendsliedrecht — can you consult with others at Animo to see if we’re ready for this.  @genaris — good to go?

andrewwhitehead:
One of the transient dependencies updated its minimum Rust version, updated the version used on CI now.

Issue #217, opened 2023-09-16
Usage of lerna 7 in CI breaks publishing script


TimoGlastra:
```
lerna ERR! ECONFIGWORKSPACES The "useWorkspaces" option has been removed. By default lerna will resolve your packages using your package manager's workspaces 
```

Should probably update to use lerna v7, but then also pin to use v7 to not break again in the future

andrewwhitehead:
I think this is resolved now, I suppose we just run the workflow again to publish the packages?

PR #219, opened 2023-09-18
ci: fix lerna version

genaris:
To address #217, I've pinned lerna to the latest version and run a `lerna repair` (as suggested in [Lerna 7.0.0 release notes](https://github.com/lerna/lerna/releases/tag/7.0.0)) to remove wrong configurations.

If this approach is OK, I think we'll need to do the same for aries-askar (already updated to lerna 7.x but config files not fixed yet) and maybe anoncreds-rs (currently using lerna 6.x).


andrewwhitehead:
Looks reasonable to me

genaris:
@andrewwhitehead what about these latest changes? I'm pinning `lerna` version directly in the GitHub action to avoid adding dev dependencies.

andrewwhitehead (approved):

PR #230, opened 2023-10-26
release dev under alpha and fix windows install

berendsliedrecht:
like: https://github.com/hyperledger/aries-askar/pull/192


TimoGlastra (dismissed):

TimoGlastra (approved):
Please also update the non-js dev versions. That way we know all dev versions are having the same features / capabilities etc..

berendsliedrecht:
> Please also update the non-js dev versions. That way we know all dev versions are having the same features / capabilities etc..

I thought from the beginning we said that version sync was not smart / not feasible, which I still agree with. Also, incrementing versions of the Python (and possibly rust lib) won't change anything in feature parity and they are on version `v0.4.0`. Maybe I misunderstood your comment, but for me all wrappers / native libs versions live separately and should not be related to each other (especially for dev).

TimoGlastra:
>  thought from the beginning we said that version sync was not smart / not feasible, which I still agree with

This is mainly for the specific version (0.1.0/0.2.0). But once we are working with a .dev version in the rust library, we do get benefit from at least keeping the dev versions in sync. It's just weird to have 0.2.0-dev.4 depends on 0.2.0-dev.2 native library. 

So until we're out of dev, I think we should keep the dev-<version> in sync

TimoGlastra:
Oh wait I understand now. The Indy VDR native library is not in dev. Then it's fine yes 👍 

My comment was mainly about if the native library has a dev version. But that's currently only the case for AnonCreds

PR #243, opened 2023-12-06
Add genesis transactions caching

andrewwhitehead:
This PR adds a new FFI method (and Python binding) to set the cache directory to use. When the cache directory is set, the library will automatically store the latest transactions for a given root hash and fetch them when creating a new pool. This means that as long as the same genesis transactions are passed in (via a string or file), it should be able to find the latest copy.

After the initial refresh of a pool, or if the transactions are cached, the library will also use a consensus `GET_TXN` request to check the latest pool state. This can be faster than the traditional status request as it will short-circuit after a single valid state proof, instead of attempting to contact all of the nodes.

### DIFF
--- a/indy-vdr-proxy/src/handlers.rs
+++ b/indy-vdr-proxy/src/handlers.rs
@@ -246,7 +246,7 @@ fn http_status_msg<T: std::fmt::Display>(code: StatusCode, msg: T) -> VdrResult<
 }
 
 async fn get_pool_genesis<T: Pool>(pool: &T) -> VdrResult<ResponseType> {
-    let txns = pool.get_json_transactions()?;
+    let txns = pool.get_transactions().encode_json()?;
     Ok(ResponseType::Genesis(txns.join("\n")))
 }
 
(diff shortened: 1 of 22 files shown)


swcurran:
Andrew — could you provide details on how this works?

Presumably we would need a JavaScript wrapper update as well.

swcurran:
@cvarjao @wadeking98 — FYI.

@berendsliedrecht — any chance you could make the update to the JavaScript wrapper?  This would be good to get into the AFJ.

swcurran:
Nice!!  It sure doesn’t explain 2 minutes for an issue credentialm but it certainly is way better!

I was wondering if we could have a minimalist script that use Indy VDR and runs some sort of reading test script against different ledgers.  It would be really nice to have that so that it could be run when we want to investigate an issue with a given ledger.  And it could demonstrate “best practices” for a caller using Indy VDR.

What do you think?

andrewwhitehead:
@swcurran Yes, that sounds useful. There's some integration tests in this repo but they might need to be generalized to test against a random ledger. It might be a bit limited given that we don't know there are any registered objects of different types, like revocation registries, but it can certainly test connectivity and fetching transactions (and optionally writing as well).

swcurran (approved):

PR #244, opened 2023-12-06
genesis cache js

berendsliedrecht:
Builds on top of #243 

- Updated c headers
- Added `set_cache_directory` to node.js and react native


swcurran (approved):
Based on @TimoGlastra ’s review

PR #245, opened 2023-12-07
Update setup.py

tnkhanh:
Not sure if this is correct but I think it makes more sense
@andrewwhitehead


andrewwhitehead (approved):
Thanks!

PR #252, opened 2024-01-18
feat: transaction caching

wadeking98:
Added functionality to cache read requests to the ledger.
I've provided added functionality to store the cache in memory or on disk in either a TTL or a LRU style cache. I've left the implementation extendable so users can easily implement storing the cache in a db, redis, whatever.

## Demo: 
The cache will keep values as valid for 24 hours, add the `--use-cache` flag to use the caching functionality. If the cache path isn't supplied then it will use a cache in memory instead of on the file system. If cache size is not specified then it will use a cache with size 1000
`./target/debug/indy-vdr-proxy -g ./genesis.txn --use-cache --cache-size 1024 --cache-path /dev/shm/vdr-cache -p 3000`
### Cache:
cache

### No Cache:
no-cache

  
  
With caching enabled on a request to lookup a schema, it takes under 32 milliseconds and is sometimes as fast as 7 milliseconds. Compared with when no cache is enabled, the same request will take over 135 milliseconds to almost a second

### DIFF
--- a/indy-vdr-proxy/Cargo.toml
+++ b/indy-vdr-proxy/Cargo.toml
@@ -34,6 +34,7 @@ serde_json = "1.0"
 tokio = { version = "1.0", features = ["macros", "rt-multi-thread", "signal"] }
 tokio-rustls = { version = "0.24", optional = true }
 url = "2.2.2"
+sled = "0.34.7"
 
 [target.'cfg(unix)'.dependencies]
 hyper-unix-connector = "0.2"
--- a/indy-vdr-proxy/Cargo.toml
+++ b/indy-vdr-proxy/Cargo.toml
@@ -34,7 +34,6 @@ serde_json = "1.0"
 tokio = { version = "1.0", features = ["macros", "rt-multi-thread", "signal"] }
 tokio-rustls = { version = "0.24", optional = true }
 url = "2.2.2"
-sled = "0.34.7"
 
 [target.'cfg(unix)'.dependencies]
 hyper-unix-connector = "0.2"
--- a/libindy_vdr/Cargo.toml
+++ b/libindy_vdr/Cargo.toml
@@ -62,6 +62,7 @@ url = "2.2.2"
 zmq = "0.9"
 async-trait = "0.1.77"
 async-lock = "3.3.0"
+sled = "0.34.7"
 
 [dev-dependencies]
 rstest = "0.18"
--- a/libindy_vdr/Cargo.toml
+++ b/libindy_vdr/Cargo.toml
@@ -60,8 +60,8 @@ thiserror = "1.0"
 time = { version = "=0.3.20", features = ["parsing"] }
 url = "2.2.2"
 zmq = "0.9"
-tokio = { version = "1.35.1", features = ["full"] }
 async-trait = "0.1.77"
+async-lock = "3.3.0"
 
 [dev-dependencies]
 rstest = "0.18"
--- a/libindy_vdr/Cargo.toml
+++ b/libindy_vdr/Cargo.toml
@@ -61,6 +61,7 @@ time = { version = "=0.3.20", features = ["parsing"] }
 url = "2.2.2"
 zmq = "0.9"
 tokio = { version = "1.35.1", features = ["full"] }
+async-trait = "0.1.77"
 
 [dev-dependencies]
 rstest = "0.18"
--- a/libindy_vdr/Cargo.toml
+++ b/libindy_vdr/Cargo.toml
@@ -60,6 +60,7 @@ thiserror = "1.0"
 time = { version = "=0.3.20", features = ["parsing"] }
 url = "2.2.2"
 zmq = "0.9"
+tokio = { version = "1.35.1", features = ["full"] }
 
 [dev-dependencies]
 rstest = "0.18"
--- a/libindy_vdr/include/libindy_vdr.h
+++ b/libindy_vdr/include/libindy_vdr.h
@@ -481,7 +481,7 @@ ErrorCode indy_vdr_resolve(PoolHandle pool_handle,
 
 ErrorCode indy_vdr_set_cache_directory(FfiStr path);
 
-ErrorCode indy_vdr_set_ledger_txn_cache(size_t capacity, c_ulong expiry_offset_ms, FfiStr path);
+ErrorCode indy_vdr_set_ledger_txn_cache(int32_t capacity, int64_t expiry_offset_ms, FfiStr path);
 
 ErrorCode indy_vdr_set_config(FfiStr config);
 
--- a/libindy_vdr/include/libindy_vdr.h
+++ b/libindy_vdr/include/libindy_vdr.h
@@ -481,6 +481,10 @@ ErrorCode indy_vdr_resolve(PoolHandle pool_handle,
 
 ErrorCode indy_vdr_set_cache_directory(FfiStr path);
 
+ErrorCode indy_vdr_set_ledger_txn_cache(size_t capacity, c_ulong expiry_offset_ms);
+
+ErrorCode indy_vdr_set_ledger_txn_fs_cache(size_t capacity, c_ulong expiry_offset_ms, FfiStr path);
+
 ErrorCode indy_vdr_set_config(FfiStr config);
 
 ErrorCode indy_vdr_set_default_logger(void);
--- a/libindy_vdr/include/libindy_vdr.h
+++ b/libindy_vdr/include/libindy_vdr.h
@@ -481,9 +481,7 @@ ErrorCode indy_vdr_resolve(PoolHandle pool_handle,
 
 ErrorCode indy_vdr_set_cache_directory(FfiStr path);
 
-ErrorCode indy_vdr_set_ledger_txn_cache(size_t capacity, c_ulong expiry_offset_ms);
-
-ErrorCode indy_vdr_set_ledger_txn_fs_cache(size_t capacity, c_ulong expiry_offset_ms, FfiStr path);
+ErrorCode indy_vdr_set_ledger_txn_cache(size_t capacity, c_ulong expiry_offset_ms, FfiStr path);
 
 ErrorCode indy_vdr_set_config(FfiStr config);
 
--- a/libindy_vdr/src/common/error.rs
+++ b/libindy_vdr/src/common/error.rs
@@ -130,6 +130,12 @@ impl From<zmq::Error> for VdrError {
     }
 }
 
+impl From<sled::Error> for VdrError {
+    fn from(err: sled::Error) -> VdrError {
+        VdrError::new(VdrErrorKind::FileSystem, None, Some(Box::new(err)))
+    }
+}
+
 impl<M> From<(VdrErrorKind, M)> for VdrError
 where
     M: fmt::Display + Send + Sync + 'static,
--- a/libindy_vdr/src/ffi/mod.rs
+++ b/libindy_vdr/src/ffi/mod.rs
@@ -15,11 +15,15 @@ mod resolver;
 
 use crate::common::error::prelude::*;
 use crate::config::{PoolConfig, LIB_VERSION};
+use crate::pool::cache::{
+    storage::{new_fs_ordered_store, OrderedHashMap},
+    strategy::CacheStrategyTTL,
+};
 use crate::pool::{FilesystemCache, PoolTransactionsCache, ProtocolVersion};
 use crate::utils::Validatable;
 
 use self::error::{set_last_error, ErrorCode};
-use self::pool::{LegerCacheConfig, LEDGER_TXN_CACHE_CONFIG, POOL_CACHE, POOL_CONFIG};
+use self::pool::{LEDGER_CACHE_STRATEGY, POOL_CACHE, POOL_CONFIG};
 
 pub type CallbackId = i64;
 
@@ -76,7 +80,7 @@ pub extern "C" fn indy_vdr_set_cache_directory(path: FfiStr) -> ErrorCode {
 pub extern "C" fn indy_vdr_set_ledger_txn_cache(capacity: usize, expire_offset: u64) -> ErrorCode {
     catch_err! {
         debug!("Setting pool ledger transactions cache: capacity={}, expire_offset={}", capacity, expire_offset);
-        *write_lock!(LEDGER_TXN_CACHE_CONFIG)? = Some(LegerCacheConfig::new(capacity, expire_offset.into(), None));
+        *write_lock!(LEDGER_CACHE_STRATEGY)? = Some(Arc::new(CacheStrategyTTL::new(capacity, expire_offset.into(), None, None)));
         Ok(ErrorCode::Success)
     }
 }
@@ -89,7 +93,8 @@ pub extern "C" fn indy_vdr_set_ledger_txn_fs_cache(
 ) -> ErrorCode {
     catch_err! {
         debug!("Setting pool ledger transactions cache: capacity={}, expire_offset={}", capacity, expire_offset);
-        *write_lock!(LEDGER_TXN_CACHE_CONFIG)? = Some(LegerCacheConfig::new(capacity, expire_offset.into(), Some(path.into_string())));
+        let store = OrderedHashMap::new(new_fs_ordered_store(path.into())?);
+        *write_lock!(LEDGER_CACHE_STRATEGY)? = Some(Arc::new(CacheStrategyTTL::new(capacity, expire_offset.into(), Some(store), None)));
         Ok(ErrorCode::Success)
     }
 }
--- a/libindy_vdr/src/ffi/mod.rs
+++ b/libindy_vdr/src/ffi/mod.rs
@@ -15,11 +15,14 @@ mod resolver;
 
 use crate::common::error::prelude::*;
 use crate::config::{PoolConfig, LIB_VERSION};
+use crate::pool::cache::storage::{new_fs_ordered_store, OrderedHashMap};
+use crate::pool::cache::strategy::CacheStrategyTTL;
+use crate::pool::cache::Cache;
 use crate::pool::{FilesystemCache, PoolTransactionsCache, ProtocolVersion};
 use crate::utils::Validatable;
 
 use self::error::{set_last_error, ErrorCode};
-use self::pool::{POOL_CACHE, POOL_CONFIG};
+use self::pool::{LEDGER_TXN_CACHE, POOL_CACHE, POOL_CONFIG};
 
 pub type CallbackId = i64;
 
@@ -68,6 +71,31 @@ pub extern "C" fn indy_vdr_set_cache_directory(path: FfiStr) -> ErrorCode {
             None
         };
         *write_lock!(POOL_CACHE)? = cache;
+        Ok(ErrorCode::Success)
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn indy_vdr_set_ledger_txn_cache(capacity: usize, expire_offset: u64) -> ErrorCode {
+    catch_err! {
+        debug!("Setting pool ledger transactions cache: capacity={}, expire_offset={}", capacity, expire_offset);
+        let cache = Cache::new(CacheStrategyTTL::new(capacity, expire_offset.into(), None, None));
+        *write_lock!(LEDGER_TXN_CACHE)? = Some(cache);
+        Ok(ErrorCode::Success)
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn indy_vdr_set_ledger_txn_fs_cache(
+    capacity: usize,
+    expire_offset: u64,
+    path: FfiStr,
+) -> ErrorCode {
+    catch_err! {
+        debug!("Setting pool ledger transactions cache: capacity={}, expire_offset={}", capacity, expire_offset);
+        let store = OrderedHashMap::new(new_fs_ordered_store(path.into())?);
+        let cache = Cache::new(CacheStrategyTTL::new(capacity, expire_offset.into(), Some(store), None));
+        *write_lock!(LEDGER_TXN_CACHE)? = Some(cache);
         Ok(ErrorCode::Success)
     }
 }
--- a/libindy_vdr/src/ffi/mod.rs
+++ b/libindy_vdr/src/ffi/mod.rs
@@ -79,8 +79,8 @@ pub extern "C" fn indy_vdr_set_cache_directory(path: FfiStr) -> ErrorCode {
 
 #[no_mangle]
 pub extern "C" fn indy_vdr_set_ledger_txn_cache(
-    capacity: usize,
-    expire_offset: u64,
+    capacity: i32,
+    expire_offset: i64,
     path_opt: FfiStr,
 ) -> ErrorCode {
     catch_err! {
@@ -90,7 +90,7 @@ pub extern "C" fn indy_vdr_set_ledger_txn_cache(
             path => OrderedHashMap::new(new_fs_ordered_store(path.into())?),
         };
 
-        *write_lock!(LEDGER_CACHE_STRATEGY)? = Some(Arc::new(CacheStrategyTTL::new(capacity, expire_offset.into(), Some(store), None)));
+        *write_lock!(LEDGER_CACHE_STRATEGY)? = Some(Arc::new(CacheStrategyTTL::new(capacity.try_into().ok().unwrap_or_default(), expire_offset.try_into().ok().unwrap_or_default(), Some(store), None)));
         Ok(ErrorCode::Success)
     }
 }
--- a/libindy_vdr/src/ffi/mod.rs
+++ b/libindy_vdr/src/ffi/mod.rs
@@ -85,10 +85,9 @@ pub extern "C" fn indy_vdr_set_ledger_txn_cache(
 ) -> ErrorCode {
     catch_err! {
         debug!("Setting pool ledger transactions cache: capacity={}, expire_offset={}", capacity, expire_offset);
-        let store = match path_opt.as_opt_str() {
-            Some("") => OrderedHashMap::new(new_mem_ordered_store()),
-            Some(path) => OrderedHashMap::new(new_fs_ordered_store(path.into())?),
-            None => OrderedHashMap::new(new_mem_ordered_store()),
+        let store = match path_opt.as_opt_str().unwrap_or_default() {
+            "" => OrderedHashMap::new(new_mem_ordered_store()),
+            path => OrderedHashMap::new(new_fs_ordered_store(path.into())?),
         };
 
         *write_lock!(LEDGER_CACHE_STRATEGY)? = Some(Arc::new(CacheStrategyTTL::new(capacity, expire_offset.into(), Some(store), None)));
--- a/libindy_vdr/src/ffi/mod.rs
+++ b/libindy_vdr/src/ffi/mod.rs
@@ -15,6 +15,7 @@ mod resolver;
 
 use crate::common::error::prelude::*;
 use crate::config::{PoolConfig, LIB_VERSION};
+use crate::pool::cache::storage::new_mem_ordered_store;
 use crate::pool::cache::{
     storage::{new_fs_ordered_store, OrderedHashMap},
     strategy::CacheStrategyTTL,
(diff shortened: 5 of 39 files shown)


libindy_vdr/src/pool/runner.rs
```diff
@@ -50,7 +52,7 @@ impl PoolRunner {
             )
             .unwrap();
             let mut thread = PoolThread::new(pool, receiver);
-            thread.run();
+            thread.run(cache);
```
andrewwhitehead, line 55: I think it would make more sense to make cache a property of `PoolThread` and leave `run()` with no arguments?

libindy_vdr/src/pool/runner.rs
```diff
@@ -217,16 +221,21 @@ impl PoolThread {
...
 }
 
-async fn _perform_refresh(pool: &LocalPool, callback: Callback<RefreshResponse>) {
-    let result = perform_refresh(pool).await;
+async fn _perform_refresh(
+    pool: &LocalPool,
+    callback: Callback<RefreshResponse>,
+    cache: Option<Cache<String, (String, RequestResultMeta)>>,
+) {
+    let result = perform_refresh(pool, cache).await;
```
andrewwhitehead, line 229: We don't want to use the cache when performing a refresh, it needs to connect to one or more nodes to see if the pool state is up to date.

libindy_vdr/Cargo.toml
```diff
@@ -60,6 +60,8 @@ thiserror = "1.0"
 time = { version = "=0.3.20", features = ["parsing"] }
 url = "2.2.2"
 zmq = "0.9"
+tokio = { version = "1.35.1", features = ["full"] }
```
andrewwhitehead, line 63: async-lock would be a much lighter dependency

swcurran:
For all but RevRegEntries, expiring the cache is not needed since the transactions are immutable. Would it make sense (is it possible?) have a fixed (maximum) sized cache with an LRU expiration policy?

Just as a matter of interest, what is the approach on RevRegEntries for finding cache hits?  It seems unlikely there would be cache hits for RevRegEntry transactions, since the combination of `from`, `to` intervals would have to be the same.

wadeking98:
> For all but RevRegEntries, expiring the cache is not needed since the transactions are immutable. Would it make sense (is it possible?) have a fixed (maximum) sized cache with an LRU expiration policy?
> 
> Just as a matter of interest, what is the approach on RevRegEntries for finding cache hits? It seems unlikely there would be cache hits for RevRegEntry transactions, since the combination of `from`, `to` intervals would have to be the same.

I like the idea of not having a time based expiration, the only thing with the LRU approach is that cache get request will have to mutate the cache as well (to mark an entry as recently used). I don't really have a problem with this but it might look a little weird for end users if the get cache requests mutate things. Still, I think it's a good idea. 

In terms of the revocation registry, I think it's still valid. For example, in the wallet the user may open a proof request multiple times, or select an alternate credential for the proof which would fetch the revocation registry entry. I think the wallet would benefit from having RevRegEntry transactions cached

andrewwhitehead:
You could add a TTL and maximum number of items, then prune the cache (removing expired items first) when adding new items.

libindy_vdr/Cargo.toml
wadeking98: Done

libindy_vdr/src/pool/runner.rs
wadeking98: Done

libindy_vdr/src/pool/runner.rs
wadeking98: Done

wadeking98:
I could add support for both and then the end user can decide which one to use, or do you think LRU should be avoided @andrewwhitehead ?

andrewwhitehead:
I think the only issue with straight LRU is that commonly used entries might never expire, which means they would never be re-fetched. Also, the get/insert/remove methods of the cache implementation don't necessarily need to take `&mut self`, because locking the Mutex gives you a mutable reference anyway.

swcurran:
I think the fetching patterns vary by object type.  Is it worth worrying about that?

Specifically:

- DIDs, ATTRIBs can be “updated” (e.g. a new instance written, and on read the latest is returned) so they do need a “moderate” TTL on them.
  - That said, for DIDs at least, we rarely need them, so reads are likely relatively rare.
- CredDefs and RevRegs don’t ever expire, so can have a very long TTL
- RevRegStatus requests (they aren’t ledger objects) must be cached using their query parameters (from, to) and as such should have a short TTL, as getting them out of the cache fairly quickly is probably a good idea.  Every time a new presentation is generated, the query is likely to be different.

wadeking98:
> I think the only issue with straight LRU is that commonly used entries might never expire, which means they would never be re-fetched. Also, the get/insert/remove methods of the cache implementation don't necessarily need to take `&mut self`, because locking the Mutex gives you a mutable reference anyway.

Got it, I've left the LRU cache there just in case. I've implemented the TTL caching functionality as well. I think that I need to keep insert and remove mutable because I've replaced the mutex on the cache with a RwLock. I only had it as a mutex when the LRU cache was the only implementation we had

wadeking98:
> I think the fetching patterns vary by object type. Is it worth worrying about that?
> 
> Specifically:
> 
>     * DIDs, ATTRIBs can be “updated” (e.g. a new instance written, and on read the latest is returned) so they do need a “moderate” TTL on them.
>       
>       * That said, for DIDs at least, we rarely need them, so reads are likely relatively rare.
> 
>     * CredDefs and RevRegs don’t ever expire, so can have a very long TTL
> 
>     * RevRegStatus requests (they aren’t ledger objects) must be cached using their query parameters (from, to) and as such should have a short TTL, as getting them out of the cache fairly quickly is probably a good idea.  Every time a new presentation is generated, the query is likely to be different.

I think this is a good idea... but I think may be big/complex enough to deserve it's own PR

andrewwhitehead:
> I think that I need to keep insert and remove mutable because I've replaced the mutex on the cache with a RwLock

The same is true for RwLock. The contents of an Arc are not generally writable, except in cases like this where a write lock is obtained.

If the TTL varies by transaction type then it might be best for the caller to specify what that is in each request? Although we could also add some defaults.

andrewwhitehead:
One important case to test would be a GET_TXN for an undefined (not yet published) sequence number. This can return a response with a 'null' value which we do not want to cache.

wadeking98:
> > I think that I need to keep insert and remove mutable because I've replaced the mutex on the cache with a RwLock
> 
> The same is true for RwLock. The contents of an Arc are not generally writable, except in cases like this where a write lock is obtained.
> 
> If the TTL varies by transaction type then it might be best for the caller to specify what that is in each request? Although we could also add some defaults.

I was thinking the TTL storage type could handle that, maybe in the `new()` function the user passes in a default ttl along with a list of enums that match the request type with an associated type of u64 or something that determines the ttl for that request type

wadeking98:
@swcurran @andrewwhitehead I've added functionality to optionally insert with custom ttl/lru index. The cache itself should be strategy agnostic (ie: doesn't care if lru or ttl is being used), so I had to add each caching strategy to an enum that has their config as an associated type, and then pass in that enum with the ttl/lru value when inserting.  Example here: https://github.com/hyperledger/indy-vdr/blob/8838b0541f49d1d30339341b99b5e14919f899f1/libindy_vdr/src/pool/cache/strategy.rs#L421

wadeking98:
> Cool. I’ll leave it to @andrewwhitehead to review the code.
> 
> What has to be done in AFJ/ACA-Py/Aries VCX to make use of this? Perhaps you could add a ticket to AFJ to outline what is needed, and then I can do the same for ACA-Py?

I will still probably need some help from andrew understanding how to export it to the FFIs and adding it to the wrappers, but after that we'll be ready to use it in afj, aca-py, etc

libindy_vdr/src/pool/cache/storage.rs
```diff
@@ -0,0 +1,219 @@
...
+        let (lookup, _) = &self.0;
+        lookup.len()
+    }
+    pub fn get(&self, key: &K) -> Option<&(O, V)> {
+        let (lookup, _) = &self.0;
+        lookup.get(key)
+    }
+    fn get_key_value(
+        &self,
+        selector: Box<
```
andrewwhitehead, line 131: I think this can take an `impl FnOnce` rather than require boxing?

indy-vdr-proxy/Cargo.toml
```diff
@@ -34,6 +34,7 @@ serde_json = "1.0"
 tokio = { version = "1.0", features = ["macros", "rt-multi-thread", "signal"] }
 tokio-rustls = { version = "0.24", optional = true }
 url = "2.2.2"
+sled = "0.34.7"
```
andrewwhitehead, line 37: It might be nice to expose a convenience wrapper in the indy_vdr crate so that the proxy doesn't also have to depend directly on sled.

libindy_vdr/src/pool/cache/strategy.rs
```diff
@@ -0,0 +1,442 @@
...
+            let removal_key = self.store.get_first_key_value().map(|(k, _, _)| k.clone());
+            if let Some(removal_key) = removal_key {
+                self.store.remove(&removal_key);
+            }
+        };
+
+        let exp_offset = match config {
+            Some(opt_vec) => opt_vec
+                .iter()
+                .find(|opt| match opt {
```
andrewwhitehead, line 105: I think you can use:

```
.find_map(|opt| match opt {
 CacheStrategyConfig::TTL(ttl) => Some(ttl),
 _ => None,
})
```

libindy_vdr/src/pool/helpers.rs
```diff
@@ -19,12 +20,13 @@ use crate::utils::base58;
 /// Perform a pool ledger status request to see if catchup is required
 pub async fn perform_pool_status_request<T: Pool>(
     pool: &T,
+    cache: Option<Cache<String, (String, RequestResultMeta)>>,
```
andrewwhitehead, line 23: No cache needed here

libindy_vdr/src/pool/cache/strategy.rs
wadeking98: I made it accept a list because the config has to be supplied at the cache level which is strategy agnostic. We don't know what cache strategy is being used under the hood and the user may want it to be at the back of the queue if it's using LRU, or they may want it to expire quickly if it's using TTL. In which case they'd supply a config option for both the LRU and TTL caching strategies

wadeking98:
@andrewwhitehead I've implemented FFI to the best of my knowledge. I saw you implemented a different cache for the pool refresh, I wasn't sure if you wanted the new cache to replace that or not so I kept them separate. I could also use a better name for the FFIs to set the cache right now they're along the lines of `indy_vdr_set_ledger_txn_cache` I wasn't sure if I should change it to something like `indy_vdr_set_runner_cache` or something like that

libindy_vdr/src/pool/builder.rs
```diff
@@ -15,16 +17,22 @@ pub struct PoolBuilder {
...
+    cache: Option<Cache<String, (String, RequestResultMeta)>>,
 }
 
 impl PoolBuilder {
     /// Create a new `PoolBuilder` instance.
-    pub fn new(config: PoolConfig, transactions: PoolTransactions) -> Self {
+    pub fn new(
+        config: PoolConfig,
+        transactions: PoolTransactions,
+        cache: Option<Cache<String, (String, RequestResultMeta)>>,
```
andrewwhitehead, line 28: It's confusing to me how PoolBuilder ignores the cache parameter unless it's used to create a PoolRunner, maybe that should just be a parameter of the into_runner method?

libindy_vdr/src/ffi/pool.rs
```diff
@@ -102,7 +107,7 @@ fn handle_pool_refresh(
             cache.update(&init_txns, latest_txns)?;
         }
         if let Some(new_txns) = new_txns {
-            let runner = PoolBuilder::new(config, new_txns).node_weights(node_weights).refreshed(true).into_runner()?;
+            let runner = PoolBuilder::new(config, new_txns, None).node_weights(node_weights).refreshed(true).into_runner()?;
```
andrewwhitehead, line 110: I think this means that after a successful pool refresh (with new pool transactions), the new pool runner will no longer have a cache?

libindy_vdr/src/pool/builder.rs
wadeking98: ya that's a better idea, done

andrewwhitehead:
I think we may be better off with one cache strategy rather than TTL and LRU as alternatives. For simplicity, but also because:

1. The cache must always be able to restrict its own size (ie. drop LRU entries) to avoid unlimited growth
2. The cache must support expiry times (TTL) in order to avoid returning long outdated records for mutable transaction types like DIDs

If each LRU cache entry also supports an optional expiry time, then the get() operation can simply return None if the cached entry has expired. We could also do automatic compaction to remove expired entries in one pass.

When fetching a transaction, it should also be possible to pass in a maximum TTL, putting additional bounds on the expiry time of the retrieved record (if any). I don't really see the benefit of overriding the LRU cache position when storing cached records.

wadeking98:
Okay got it, so should I remove the LRU strategy, or should I modify it to support TTL so we'd have a ttl/lru hybrid?

andrewwhitehead:
Yes, I think I would modify the LRU one to support automatic expiry (TTL).

wadeking98:
got it, and which strategy should I expose to the FFI? should I expose both?

andrewwhitehead:
I don't think we want to support multiple strategies, just the hybrid. @swcurran ?

libindy_vdr/src/ffi/pool.rs
```diff
@@ -40,6 +41,9 @@ pub static POOLS: Lazy<RwLock<BTreeMap<PoolHandle, PoolInstance>>> =
 pub static POOL_CACHE: Lazy<RwLock<Option<Arc<dyn PoolTransactionsCache>>>> =
     Lazy::new(|| RwLock::new(Some(Arc::new(InMemoryCache::new()))));
 
+pub static LEDGER_TXN_CACHE: Lazy<RwLock<Option<Cache<String, (String, RequestResultMeta)>>>> =
```
andrewwhitehead, line 44: I think there may be an issue with having one global cache. Clients often connect to multiple ledgers, and the caches need to be distinct, so that a schema ID from one ledger doesn't resolve via the cache when requesting from another.

wadeking98: the cache key includes the ledger ID:
image

wadeking98: I'm using sha256 to hash the key so we can save space in the cache. We shouldn't have any conflict caching for multiple ledgers in the same cache

wadeking98:
@andrewwhitehead I've merged the TTL and LRU strategies. I kept the old TTL name since it's mostly a TTL cache

libindy_vdr/src/ffi/pool.rs
andrewwhitehead: I believe that `ledgerId` identifies the plenum 'subledger': 0 is the pool, 1 is the domain ledger for most commonly used transactions, 2 is the config ledger. It is good to include that in the cache key, but it won't differentiate responses from (say) CANdy and Sovrin.

libindy_vdr/src/pool/cache/strategy.rs
```diff
@@ -0,0 +1,199 @@
...
+            .unwrap();
+        let storage: OrderedHashMap<String, u128, TTLCacheItem<String>> = OrderedHashMap::new(tree);
+        let fs_cache = Cache::new(CacheStrategyTTL::new(2, 5, Some(storage), None));
+        let caches = vec![cache, fs_cache];
+        block_on(async {
+            for cache in caches {
+                cache
+                    .insert("key".to_string(), "value".to_string(), None)
+                    .await;
+                thread::sleep(std::time::Duration::from_millis(1));
```
andrewwhitehead, line 149: I'm not sure why these sleeps are necessary, is the cache not updated by the time insert completes?

wadeking98: They're not really necessary, I just put them there for testing because otherwise all the values in the BtreeMap or sled storage would be keyed under the same millisecond

libindy_vdr/src/ffi/pool.rs
wadeking98: how did you differentiate between ledgers when you created the pool cache? or did you not need to?

andrewwhitehead: It's not needed there, the set of genesis transactions (the cached values) are enough to differentiate them.

wadeking98: could we use the pool merkle tree as part of the key? that way it would differentiate between each pool? Having a global cache is nice because then we just have one cache location on the file system, I'd rather not make the user maintain multiple db files for each ledger they use

andrewwhitehead: You could, but then you'd lose the cache when the pool transactions were updated. I think even if you picked like, the first 4 transactions you would risk strange effects when connecting to two ledgers that had forked off of the same base (although that is somewhat theoretical).

wadeking98: hmm, is there any way to consistently ID a pool across pool refreshes/updates?

wadeking98: if not I could add a pool id to the type that we could leverage in our caching

wadeking98:
@andrewwhitehead I've added functionality to have one cache per ledger to avoid conflicts. If the client connects to two ledgers that have the same initial transactions and uses a filesystem cache then there is still a potential for cache conflicts. But like you said that case may be somewhat theoretical since I'm not sure how many different ledgers are likely to have the same genesis transactions

libindy_vdr/src/ffi/mod.rs
```diff
@@ -72,6 +76,29 @@ pub extern "C" fn indy_vdr_set_cache_directory(path: FfiStr) -> ErrorCode {
...
+        Ok(ErrorCode::Success)
+    }
+}
+
+#[no_mangle]
+pub extern "C" fn indy_vdr_set_ledger_txn_fs_cache(
+    capacity: usize,
+    expire_offset: u64,
+    path: FfiStr,
+) -> ErrorCode {
```
andrewwhitehead, line 93: Could we combine these two methods and just make the path an optional parameter, using the in-memory cache if it's empty or null?

libindy_vdr/src/ffi/pool.rs
```diff
@@ -102,7 +110,9 @@ fn handle_pool_refresh(
             cache.update(&init_txns, latest_txns)?;
         }
         if let Some(new_txns) = new_txns {
-            let runner = PoolBuilder::new(config, new_txns).node_weights(node_weights).refreshed(true).into_runner()?;
+            // set this cache with unique key prefix
+            let txn_cache = read_lock!(LEDGER_CACHE_STRATEGY)?.as_ref().map(|s| Cache::new(s.clone(), init_txns.root_hash_base58().ok()));
```
andrewwhitehead, line 114: I believe it should be using the hash of the new transactions here, not the initial ones

libindy_vdr/src/pool/cache/mod.rs
```diff
@@ -0,0 +1,61 @@
...
+    key_prefix: Option<K>,
+}
+
+impl<K: Display + 'static, V: 'static> Cache<K, V> {
+    fn full_key(&self, key: &K) -> String {
+        match &self.key_prefix {
+            Some(prefix) => format!("{}{}", prefix, key),
+            None => key.to_string(),
+        }
+    }
```
andrewwhitehead, line 28: I think rustfmt needs to be run on this file and the next? It seems odd not having a blank line between methods.

libindy_vdr/src/pool/cache/storage.rs
```diff
@@ -0,0 +1,227 @@
...
+        BTreeMap::remove(self, key)
+    }
+    fn entries(&self) -> Box<dyn Iterator<Item = (O, V)> + '_> {
+        Box::new(self.iter().map(|(o, v)| (o.clone(), v.clone())))
+    }
+}
+/// A hashmap that also maintains a BTreeMap of keys ordered by a given value
+/// This is useful for structures that need fast O(1) lookups, but also need to evict the oldest or least recently used entries
+/// The Ordered Store must contain both the keys and values for persistence
+pub struct OrderedHashMap<K, O, V>(
```
andrewwhitehead, line 105: I'm not sure why this is a tuple struct with a 2-tuple as its only member, instead of having two named fields?

libindy_vdr/src/ffi/pool.rs
wadeking98: That would change the ID on a pool refresh though wouldn't it? Unless that's what we want. I went with the init_txns assuming those were the genesis transactions and were less likely to change. Additionally the pool cache seems to be using them already as a key so I figured we could use them as a (somewhat) stable key prefix for the txn cache

andrewwhitehead:
Getting pretty close now I think. The solution for handling multiple ledgers should work, it's not the end of the world if the cache gets expired when the pool transactions are changed.

We may want to swap out `sled` later as it seems to be going without maintenance at the moment, but it's fine for now. `fjall` could be a good option? I'd like to find a key-value DB with random element access (or add that feature) so that this could be implemented for the cache eviction: https://danluu.com/2choices-eviction/

libindy_vdr/src/ffi/mod.rs
wadeking98: done

libindy_vdr/src/pool/cache/mod.rs
wadeking98: rust fmt didn't seem to complain, I added the linebreaks anyway though

wrappers/javascript/indy-vdr-nodejs/src/NodeJSIndyVdr.ts
```diff
@@ -141,6 +141,11 @@ export class NodeJSIndyVdr implements IndyVdr {
     this.handleError(this.nativeIndyVdr.indy_vdr_set_cache_directory(path))
   }
 
+  public setLedgerTxnCache(options: { capacity: number; expiry_offset_ms: number; path: string }): void {
```
wadeking98, line 144: the path argument here could be optional `path?: string` but I'm not sure how well that would cross the Typescript => C => Rust function interfaces. Thought? @andrewwhitehead

libindy_vdr/src/ffi/pool.rs
wadeking98: I changed it to use the new_txns anyway but I figured I'd ask

andrewwhitehead: ACA-Py does its own caching, so in that case the genesis transactions will change over time anyway. If there are two successful refreshes, then it would also have different 'initial' transactions. I think for consistency and avoiding these edge cases we can just stick with the latest.

libindy_vdr/src/ffi/mod.rs
```diff
@@ -72,6 +77,25 @@ pub extern "C" fn indy_vdr_set_cache_directory(path: FfiStr) -> ErrorCode {
...
+#[no_mangle]
+pub extern "C" fn indy_vdr_set_ledger_txn_cache(
+    capacity: usize,
+    expire_offset: u64,
+    path_opt: FfiStr,
+) -> ErrorCode {
+    catch_err! {
+        debug!("Setting pool ledger transactions cache: capacity={}, expire_offset={}", capacity, expire_offset);
+        let store = match path_opt.as_opt_str() {
+            Some("") => OrderedHashMap::new(new_mem_ordered_store()),
```
andrewwhitehead, line 89:
```suggestion
            None | Some("") => OrderedHashMap::new(new_mem_ordered_store()),
```
or you could match `path_opt.as_opt_str().unwrap_or_default()`

libindy_vdr/include/libindy_vdr.h
```diff
@@ -481,6 +481,8 @@ ErrorCode indy_vdr_resolve(PoolHandle pool_handle,
 
 ErrorCode indy_vdr_set_cache_directory(FfiStr path);
 
+ErrorCode indy_vdr_set_ledger_txn_cache(size_t capacity, c_ulong expiry_offset_ms, FfiStr path);
```
berendsliedrecht, line 484: For all other functions we use `int64_t` or `int32_t`. Could `size_t` (`usize`) and `c_ulong` (`u64`) be switched to `i32` and `i64` respectively?

libindy_vdr/src/common/error.rs
```diff
@@ -130,6 +130,12 @@ impl From<zmq::Error> for VdrError {
     }
 }
 
+impl From<sled::Error> for VdrError {
```
berendsliedrecht, line 133: Can leave this as-is for now, but I think we can use `#[from(sled::Error)]` from `thiserror` and it will derive it properly.

libindy_vdr/src/pool/cache/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+#[async_trait]
+pub trait CacheStrategy<K, V>: Send + Sync + 'static {
+    async fn get(&self, key: &K) -> Option<V>;
+
+    async fn remove(&mut self, key: &K) -> Option<V>;
+
+    async fn insert(&mut self, key: K, value: V, custom_exp_offset: Option<u128>) -> Option<V>;
+}
+
+pub struct Cache<K, V> {
```
berendsliedrecht, line 17: Would it make sense to require `K` to implement `Display` here as well?

libindy_vdr/src/pool/cache/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+        let full_key = self.full_key(key);
+        self.storage.read().await.get(&full_key).await
+    }
+
+    pub async fn remove(&self, key: &K) -> Option<V> {
+        let full_key = self.full_key(key);
+        self.storage.write().await.remove(&full_key).await
+    }
+
+    pub async fn insert(&self, key: K, value: V, custom_exp_offset: Option<u128>) -> Option<V> {
```
berendsliedrecht, line 47: Have never used it myself but do we know the support for `u128`? Is this something that LLVM fixes for us?

libindy_vdr/src/pool/requests/prepared_request.rs
```diff
@@ -80,6 +81,18 @@ impl PreparedRequest {
         }
     }
 
+    pub fn get_cache_key(&self) -> VdrResult<String> {
+        let mut req_json = self.req_json.clone();
+        let req_map = req_json
+            .as_object_mut()
+            .ok_or_else(|| input_err("Invalid request JSON"))?;
+        req_map.remove("reqId");
+        req_map.remove("signature");
```
berendsliedrecht, line 90: This still refers to the `req_json`, right? Otherwise the removes don't do anything as `req_json` is being used after.

wrappers/javascript/indy-vdr-nodejs/src/NodeJSIndyVdr.ts
berendsliedrecht: undefined will be converted into a nullptr. Should be fine to keep the type at `path?: string`.

wrappers/javascript/indy-vdr-react-native/cpp/indyVdr.cpp
```diff
@@ -32,6 +32,16 @@ jsi::Value setCacheDirectory(jsi::Runtime &rt, jsi::Object options) {
   return createReturnValue(rt, code, nullptr);
 };
 
+jsi::Value setLedgerTxnCache(jsi::Runtime &rt, jsi::Object options) {
+  auto capacity = jsiToValue<size_t>(rt, options, "capacity");
+  auto expiry_offset_ms = jsiToValue<c_ulong>(rt, options, "expiry_offset_ms");
+  auto path = jsiToValue<std::string>(rt, options, "path");
```
berendsliedrecht, line 38: if path is optional the following can be used:

```suggestion
  auto path = jsiToValue<std::string>(rt, options, "path", true);
```

But make sure that before you call `path.c_str()` you first check whether its a nullptr or string. There should be some examples in other funtions to check it from.

wrappers/javascript/indy-vdr-react-native/src/NativeBindings.ts
```diff
@@ -13,6 +13,8 @@ export interface NativeBindings {
 
   setCacheDirectory(options: { path: string }): ReturnObject<never>
 
+  setLedgerTxnCache(options: { capacity: number; expiry_offset_ms: number; path: string }): ReturnObject<never>
```
berendsliedrecht, line 16: All the types can be with `path?: string` for the RN wrapper as well.

libindy_vdr/include/libindy_vdr.h
wadeking98: we could, but switching to signed integers would cut the maximum capacity and expiry date in half, and it doesn't really make sense to me to support negative capacity or an expiry date that's in the past

libindy_vdr/src/pool/cache/mod.rs
wadeking98: Sure thing, it's not required by rust, but now that I look it seems like a convention in the codebase

libindy_vdr/src/pool/requests/prepared_request.rs
wadeking98: yes req_map is a mutable reference to req_json, but in a slightly different type. `as_object_mut()` returns a mutable reference to req_json in the form of a `&mut Map<String, Value>` object. Insertions and removals made to the map are still reflected in the original object. I think what's happening here is that `as_object_mut()` is returning a reference to the underlying struct that serde_json uses in it's `Value` type

wadeking98: I've tested this and can confirm that the `reqId` and `signiture` attributes are removed before making the raw key

libindy_vdr/include/libindy_vdr.h
andrewwhitehead: Signed numbers are mostly used in the FFI because (supposedly) Java handles those much better than unsigned ones.

libindy_vdr/src/pool/cache/mod.rs
wadeking98: In this thread here: https://stackoverflow.com/questions/57340308/how-does-rusts-128-bit-integer-i128-work-on-a-64-bit-system 
it looks like u128s are compiled down to LLVM and handled by multiple registers

libindy_vdr/include/libindy_vdr.h
wadeking98: okay got it, I'll change to signed

berendsliedrecht (approved):

wadeking98:
I'll need @berendsliedrecht or @andrewwhitehead to merge since I don't have write access to the repo

andrewwhitehead (approved):

PR #253, opened 2024-02-08
fix: update to indy-vdr native binary 0.4.1

TimoGlastra:
Signed-off-by: Timo Glastra <timo@animo.id>

NOTE: this is to the js-0.4.1 branch and the release for JS wrapper must be created from that branch

image

### DIFF
--- a/wrappers/javascript/indy-vdr-nodejs/package.json
+++ b/wrappers/javascript/indy-vdr-nodejs/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@hyperledger/indy-vdr-nodejs",
-  "version": "0.2.0-dev.7",
+  "version": "0.2.0",
   "license": "Apache-2.0",
   "description": "Nodejs wrapper for Indy Vdr",
   "source": "src/index",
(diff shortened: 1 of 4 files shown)


berendsliedrecht (approved):

swcurran:
Just to understand — why the separation of this from the `main` branch?  Does that create a pile of complexity?  Or more specifically — when does this branch merge into main?  I’m sure there are good reasons relating to wrappers :-) — just wondering about them

swcurran:
Found this from @TimoGlastra on Discord to answer my question on why this is being done:

> FYI -- I've created a custom js-0.4.1 branch in the Indy VDR repository as the Indy VDR release for JS was still on 0.4.0 and we need to update to 0.4.1. However with the transaction caching added there's new FFI methods which aren't availalbe in 0.4.1 binary and thus will fail if we publish from the main branch. We're at the point of creating a 0.2.0 release of JS wrapper using 0.4.1 binary, and then we can stay in sync with main again

That's what I needed.  Sounds good to me.

@wadeking98 -- are you working on the FFI/wrapper part of your transaction caching work? Or is there an open issue that someone needs to do?

PR #262, opened 2024-03-31
fix(js): convert seqNo to -1 if 0

TimoGlastra:
Indy VDR does not allow a value of `0` to be passed, and the jsiToValue uses `0` as a default value if the value wasn't provided and it's optional. Same as is done in the Node.JS wrapper we convert the value to -1 (indicating it has no value).


wrappers/javascript/indy-vdr-react-native/cpp/indyVdr.cpp
```diff
@@ -240,11 +240,13 @@ jsi::Value buildGetNymRequest(jsi::Runtime &rt, jsi::Object options) {
   auto seqNo = jsiToValue<int32_t>(rt, options, "seqNo", true);
```
berendsliedrecht, line 240: `jsiToValue` could take an optional param which sets the undefined value (-1) in this case. 

Not really required to change, but it can be a nice addition in the future if we run into this again.

berendsliedrecht (approved):

PR #265, opened 2024-04-08
chore: update wrapper to 0.2.2

genaris:
This also adds the update of ffi-napi that was made to main


swcurran:
Is this the same as a full release of Indy VDR? Seems like it is just the wrapper. If not — should we do a full release as well?

genaris:
> Is this the same as a full release of Indy VDR? Seems like it is just the wrapper. If not — should we do a full release as well?

Yes, just the wrapper. I made this PR against `js-0.4.1` branch because last week the JS wrapper 0.2.1 release was deployed from there.

Issue #268, opened 2024-04-24
Release wrappers for Indy VDR version 0.4.2


swcurran:
Indy VDR 0.4.2 was released yesterday that adds support for caching.  We'd like to get all the wrappers released, especially the one for Javascript/ReactNative.

swcurran:
@genaris -- could you do this by any chance?  It looks like you did the the previous one.

@cvarjao, @wadeking98  -- FYI.

TimoGlastra:
I don't think the Indy VDR wrappers have been updated yet for JS/React Native, hence the https://github.com/hyperledger/indy-vdr/tree/js-0.4.1 branch.

Someone would need to make sure that the wrapper for both Node.JS and React Native is up to date and working before we can go ahead with the release

cvarjao:
@TimoGlastra , did you start that branch? can we get that work rebase and merged to main first? `0.2.2` is already been using by Credo 0.5.x, so I think we should get to main.

wadeking98:
I've created a PR to grab the latest binary for the js wrappers: https://github.com/hyperledger/indy-vdr/pull/270

Issue #216, opened 2023-09-14
Update React Native wrapper with changes for 0.4.0 release


TimoGlastra:
There's been some changes to the c-callable interface since the last release, mainly related to the support for did:indy. 

Thse changes have been applied to Node.JS, but not React Native: https://github.com/hyperledger/indy-vdr/commit/f21a882b07dd2b66eb84b10bc15ad6b2aad3bc5a

There may be other changes that are not in the Node.JS wrapper as well.

TimoGlastra:
Need a way to test the React Native app simply. @berendsliedrecht do you still have the test app?

genaris:
As discussed in AFJ WG call today, for the time being we'll update the React Native wrapper and release a dev version of the wrapper that can be tested in AFJ. Once we are confident with it, we can go for the official release.

genaris:
I've started looking at it and already fixed React Native wrapper to match NodeJS in https://github.com/hyperledger/indy-vdr/pull/218, but it seems that there are a few more things that weren't updated yet into JS wrappers:

- indy_vdr_build_auth_rule_request
- indy_vdr_build_auth_rules_request
- indy_vdr_build_get_auth_rule_request
- indy_vdr_build_get_frozen_ledgers_request
- indy_vdr_build_ledgers_freeze_request
- indy_vdr_build_node_request (it is only defined in nodejs wrapper's `NativeBindings.ts`) 
- indy_vdr_build_pool_config_request (it is only defined in nodejs wrapper's `NativeBindings.ts`) 
- indy_vdr_build_pool_restart_request
- indy_vdr_build_pool_upgrade_request
- indy_vdr_resolve
- indy_vdr_dereference
- indy_vdr_string_free

TimoGlastra:
@genaris did you ever make a start at the other methods? Do you think we can release a new stable release without including these new features for now?

PR #270, opened 2024-05-08
feat: updated to fetch 0.4.2 binary

wadeking98:
Updated the javascript wrappers to use the 0.4.2 binary

### DIFF
--- a/wrappers/javascript/indy-vdr-nodejs/README.md
+++ b/wrappers/javascript/indy-vdr-nodejs/README.md
@@ -51,3 +51,13 @@ yarn test:local-build
 ```
 
 > **Note**: If you want to use this library in a cross-platform environment you need to import methods from the `@hyperledger/indy-vdr-shared` package instead. This is a platform independent package that allows to register the native bindings. The `@hyperledger/indy-vdr-nodejs` package uses this package under the hood. See the [Indy VDR Shared README](https://github.com/hyperledger/indy-vdr/tree/main/wrappers/javascript/indy-vdr-shared/README.md) for documentation on how to use this package.
+
+## Version Compatibility
+
+The JavaScript wrapper is versioned independently from the native bindings. The following table shows the compatibility between the different versions:
+
+| Indy VDR      | JavaScript Wrapper |
+| ------------- | ------------------ |
+| v0.4.0-dev.16 | v0.1.0             |
+| v0.4.1        | v0.2.0             |
+| v0.4.2        | v0.2.3             |
--- a/wrappers/javascript/indy-vdr-nodejs/README.md
+++ b/wrappers/javascript/indy-vdr-nodejs/README.md
@@ -59,5 +59,5 @@ The JavaScript wrapper is versioned independently from the native bindings. The 
 | Indy VDR      | JavaScript Wrapper |
 | ------------- | ------------------ |
 | v0.4.0-dev.16 | v0.1.0             |
-| v0.4.1        | v0.2.0             |
-| v0.4.2        | v0.2.3             |
+| v0.4.1        | v0.2.x             |
+| v0.4.2        | v0.3.0             |
--- a/wrappers/javascript/indy-vdr-react-native/cpp/turboModuleUtility.cpp
+++ b/wrappers/javascript/indy-vdr-react-native/cpp/turboModuleUtility.cpp
@@ -143,7 +143,7 @@ int64_t jsiToValue(jsi::Runtime &rt, jsi::Object &options, const char *name,
                    bool optional) {
   jsi::Value value = options.getProperty(rt, name);
   if ((value.isNull() || value.isUndefined()) && optional)
-    return 0;
+    return -1;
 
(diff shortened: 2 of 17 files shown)


wrappers/javascript/indy-vdr-nodejs/README.md
```diff
@@ -51,3 +51,13 @@ yarn test:local-build
...
+
+## Version Compatibility
+
+The JavaScript wrapper is versioned independently from the native bindings. The following table shows the compatibility between the different versions:
+
+| Indy VDR      | JavaScript Wrapper |
+| ------------- | ------------------ |
+| v0.4.0-dev.16 | v0.1.0             |
+| v0.4.1        | v0.2.0             |
+| v0.4.2        | v0.2.3             |
```
genaris, line 63: I think it will be clearer if we explicitly mention that v0.4.1 corresponds to 0.2.0, 0.2.1 and 0.2.2, or just 0.2.x and call this one (the one under development) as 0.3.0.

Actually in any case I'd prefer to move on to 0.3.0

genaris (commented):
I think it is good to put the efforts in the main branch but we just need to make sure that it works properly in both nodejs and react-native, as mentioned in https://github.com/hyperledger/indy-vdr/issues/268#issuecomment-2075643725. I didn't follow the latest releases (other than some minor fixes) but I guess the `js-0.4.1` branch was created for a compatibility issue so it would be nice to get feedback from @TimoGlastra or @berendsliedrecht.

Besides that, I remember there were several methods not implemented in the wrappers even for 0.4.1, so it would be good also to take a look if they are already there now or they need to be added before doing another official release: see https://github.com/hyperledger/indy-vdr/issues/216#issuecomment-1723649279

swcurran:
I’d love to get this into a single branch and to have unified releases of all components at the same time. Additional overhead, but pretty important!

wadeking98:
@genaris is there an easy way to test it on node and react native? from my understanding it only pulls the binaries when you run npm install from the npm repository.

wrappers/javascript/indy-vdr-react-native/cpp/turboModuleUtility.cpp
```diff
@@ -143,7 +143,7 @@ int64_t jsiToValue(jsi::Runtime &rt, jsi::Object &options, const char *name,
                    bool optional) {
   jsi::Value value = options.getProperty(rt, name);
   if ((value.isNull() || value.isUndefined()) && optional)
-    return 0;
+    return -1;
```
TimoGlastra, line 146: Is it correct to return -1 always in this case? 

In that case the above code that transforms it to -1 if it's `0` can also be removed

wrappers/javascript/indy-vdr-react-native/native/indy_vdr.node
```diff
@@ -1,3 +0,0 @@
-# This is a placeholder file to prevent node-pre-gyp from installing the indy_vdr binary when cloning
```
TimoGlastra, line 1: This needs to stay!

TimoGlastra (changes requested):

TimoGlastra:
@berendsliedrecht maybe we can also add the example app to this branch so we can more easily test changes in react native?

TimoGlastra:
> I’d love to get this into a single branch and to have unified releases of all components at the same time. Additional overhead, but pretty important!

I think that makes sense, however it means we will need to require every pull request to implement functionality for all wrappers. It's sometimes quite hard to look in which way a wrapper is outdated, as you have to walk through all commits, see the changes, update it, test the different wrappers, etc..

wrappers/javascript/indy-vdr-react-native/cpp/turboModuleUtility.cpp
berendsliedrecht: Yes, after a quick check again -1 is interpreted as `None`. 

@wadeking98 could you fix all the `0`s to `-1`?

wrappers/javascript/indy-vdr-react-native/native/indy_vdr.node
wadeking98: sorry about that, I was testing some things in nodejs and didn't mean to commit it

wrappers/javascript/indy-vdr-react-native/cpp/turboModuleUtility.cpp
wadeking98: @berendsliedrecht I've done it for int64 and int32 but I'm not sure what to do for uint 64,32,8 because it needs to return unsigned

wadeking98:
I made some changes to the binary to make it work with react native. For some reason the `#[async_trait]` macro wasn't working well with react native so I removed it and refactored the code to compensate. I'm going to split up this PR to update the binary and one to update the wrapper

TimoGlastra:
I think this PR can be closed right @wadeking98? I made #296 with updates and a new version for the JS wrapper

PR #271, opened 2024-05-10
chore: updating dependabot file to support gha, TS, JS, gradle, Go lang, cargo and docker packages

rajpalc7:


WadeBarnes (approved):
Looks good to me, but I'd like feedback from other maintainers.

berendsliedrecht (approved):
LGTM.

WadeBarnes:
@TimoGlastra, We could look into grouping the dependency updates, but I've also heard complaints from developers that that causes other issues.

One option it to use the dependabot updates as a guide and create a separate PR that updates all of the dependancies that can be updated without issue.  Merging that PR will trigger dependabot to reevaluate and close any PRs that are no longer relevant.

TimoGlastra:
Yeah the thing I don't like about dependabot is that a lot of these dependencies are dev dependencies (for javascript especially), and that you're spending a lot of time on updates that don't really improve things.

WadeBarnes:
> Yeah the thing I don't like about dependabot is that a lot of these dependencies are dev dependencies (for javascript especially), and that you're spending a lot of time on updates that don't really improve things.

It looks like that could be controlled; https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#allow

PR #288, opened 2024-05-15
fix: fixed binary for react native

wadeking98:
the async_trait macro appeared to be crashing Bifold in react native. I've refactored the code to remove the `async_trait` macro and made some changes to the code to compensate

### DIFF
--- a/libindy_vdr/Cargo.toml
+++ b/libindy_vdr/Cargo.toml
@@ -60,8 +60,6 @@ thiserror = "1.0"
 time = { version = "=0.3.20", features = ["parsing"] }
 url = "2.2.2"
 zmq = "0.9"
-async-trait = "0.1.77"
-async-lock = "3.3.0"
 sled = "0.34.7"
 
 [dev-dependencies]
--- a/libindy_vdr/src/pool/cache/mod.rs
+++ b/libindy_vdr/src/pool/cache/mod.rs
@@ -1,17 +1,17 @@
-use async_lock::RwLock;
-use async_trait::async_trait;
-use std::{fmt::Display, sync::Arc};
+use std::{
+    fmt::Display,
+    sync::{Arc, RwLock},
+};
 
 pub mod storage;
 pub mod strategy;
 
-#[async_trait]
 pub trait CacheStrategy<K, V>: Send + Sync + 'static {
-    async fn get(&self, key: &K) -> Option<V>;
+    fn get(&self, key: &K) -> Option<V>;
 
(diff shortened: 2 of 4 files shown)


wadeking98:
@genaris @TimoGlastra @andrewwhitehead could I get a quick review on this please?

wadeking98:
I'd get this error message on the build with async trait after fetching a schema definition from the ledger:
`Fatal signal 11 (SIGSEGV), code 2 (SEGV_ACCERR), fault addr 0x76d39d6fa0 in tid 15686 (mqt_js), pid 15558 (bc.gov.BCWallet)`

berendsliedrecht:
> I'd get this error message on the build with async trait after fetching a schema definition from the ledger:
> 
> `Fatal signal 11 (SIGSEGV), code 2 (SEGV_ACCERR), fault addr 0x76d39d6fa0 in tid 15686 (mqt_js), pid 15558 (bc.gov.BCWallet)`

But there is no indication of async_trait breaking anything here, correct? 

I am not aware of the technical implications of removing async_trait and moving to sync, so it might be fine. 

https://github.com/dtolnay/async-trait?tab=readme-ov-file#explanation just shows that it adds a box+pin to it, which shouldn't segfault, right?

wadeking98:
it shouldn't segfault because it doesn't in node... but for some reason it does on react native maybe because it's compiled for mobile architectures (arm64 and eabi). My hunch is that box+pin isn't getting properly translated when cross compiling so maybe things are moving around in memory when they should be pinned. That's just a hunch though, I haven't proven it. 

Either way I discovered that the async_trait wasn't needed anyway and I was able to do the same functionality using core rust libraries. So I view this PR as a kind of refactoring/code cleanup with the added bonus of fixing a bug for mobile users

andrewwhitehead:
From a quick search it looks like SEGV_ACCERR usually indicates a stack overflow, which could be due to a smaller stack being configured on the mobile builds. `async_trait` would not be the problem so much as running the cache operations within the async executor, although I don't really see why that should cause any issues. If it works then this fix seems fine.

The other build errors seem to have started with the dependabot update PR, but are likely due to a new release of some dependency (cmake?). On Darwin I see it trying to link an arm64 build of libzmq during the x64_64 build.

cvarjao:
@wadeking98, #292 is merged. Try rebasing your PR.

wadeking98:
@andrewwhitehead @cvarjao the checks completed successfully

andrewwhitehead (approved):

PR #293, opened 2024-05-29
feat: added improved gh actions

wadeking98:
Added github actions to automatically publish alpha versions to npm.
Included binaries in react native package so we have a smoother release cycle without having to wait for a binary release

This PR does not effect the usual release cycle, this only adds alpha releases. Normal major releases will still have to be triggered manually

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -427,7 +427,7 @@ jobs:
             library-x86_64-linux-android
           failOnError: false
 
-  release-alpha:
+  alpha-release:
     name: Alpha Release
     if: github.event_name == 'push'
     permissions:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -427,6 +427,57 @@ jobs:
             library-x86_64-linux-android
           failOnError: false
 
+  release-alpha:
+    name: Alpha Release
+    if: github.event_name == 'push'
+    permissions:
+      id-token: write
+    runs-on: ubuntu-latest
+    env:
+      NPM_TAG: 'alpha'
+    needs: 
+      - create-ios-xcframework
+      - create-android-library
+    defaults:
+      run:
+        working-directory: ./wrappers/javascript
+    timeout-minutes: 7
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20.x'
+          registry-url: 'https://registry.npmjs.org'
+          cache: "yarn"
+          cache-dependency-path: "wrappers/javascript"
+      - name: install
+        run: yarn install
+
+      - name: Fetch Android libraries
+        uses: actions/download-artifact@v4
+        with:
+          name: android-libraries
+          path: indy-vdr-react-native/native/mobile/android/
+
+      - name: Fetch iOS Framework
+        uses: actions/download-artifact@v4
+        with:
+          name: indy_vdr.xcframework
+          path: indy-vdr-react-native/native/mobile/ios/
+
+      # On push to main, release unstable version
+      - name: Release alpha
+        run: |
+          git update-index --assume-unchanged $(find . -type d -name node_modules -prune -o -name 'package.json' -print | tr "\n" " ")
+          #export NEXT_VERSION_BUMP=$(yarn next-version-bump)
+          export NEXT_VERSION_BUMP="major"
+          npx lerna publish --no-verify-access --no-private --loglevel=verbose --canary $NEXT_VERSION_BUMP --dist-tag $NPM_TAG --force-publish --yes
+        env:
+          NPM_TOKEN: ${{ secrets.NPM_PUBLISH }}
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH }}
+
   create-ios-android-release-asset:
     name: Create iOS and Android release assets
     runs-on: ubuntu-latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -429,7 +429,7 @@ jobs:
 
   alpha-release:
     name: Alpha Release
-    if: github.event_name == 'push'
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
     permissions:
       id-token: write
     runs-on: ubuntu-latest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -253,7 +253,7 @@ jobs:
           name: library-linux-x86_64
 
       - name: Install dependencies
-        run: yarn install
+        run: yarn install --immutable
 
       - name: Build
         run: yarn build
@@ -453,7 +453,7 @@ jobs:
           cache: "yarn"
           cache-dependency-path: "wrappers/javascript"
       - name: install
-        run: yarn install
+        run: yarn install --immutable
 
       - name: Fetch Android libraries
         uses: actions/download-artifact@v4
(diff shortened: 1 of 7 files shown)


.github/workflows/build.yml
```diff
@@ -427,6 +427,57 @@ jobs:
             library-x86_64-linux-android
           failOnError: false
 
+  alpha-release:
+    name: Alpha Release
+    if: github.event_name == 'push'
```
andrewwhitehead, line 432: I think this might need to check the branch name is `main` as well?

wadeking98: on line 9 it only listens for the push event on main, so this job will only run on main. But I do think it's a good idea to add an extra check here in case the event listener up top changes in the future

.github/workflows/build.yml
```diff
@@ -427,6 +427,57 @@ jobs:
...
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20.x'
+          registry-url: 'https://registry.npmjs.org'
+          cache: "yarn"
+          cache-dependency-path: "wrappers/javascript"
+      - name: install
+        run: yarn install
```
cvarjao, line 456: do we need `--immutable`?

cvarjao (changes requested):

.github/workflows/build.yml
```diff
@@ -427,6 +427,57 @@ jobs:
...
+    env:
+      NPM_TAG: 'alpha'
+    needs: 
+      - create-ios-xcframework
+      - create-android-library
+    defaults:
+      run:
+        working-directory: ./wrappers/javascript
+    timeout-minutes: 7
+    steps:
```
jleach, line 445: A few spaces won't change much but make make it easier for the eye to follow. Between steps?

wrappers/javascript/indy-vdr-react-native/package.json
```diff
@@ -54,12 +53,5 @@
   "peerDependencies": {
     "react": ">= 16",
     "react-native": ">= 0.66.0"
-  },
-  "binary": {
```
berendsliedrecht, line 58: Won't this break the non-alpha releases? My preference would really be to fetch the libraries from gh in a post install (what node-pre-gyp does), but if there are benefits I am okay with it.

cvarjao: we can go back to fetch when/if we can figure out publishing alpha releases of the rust libraries. I was thinking on maybe publishing alphas builds to GitHub Package Registry (maybe as maven artifacts)

cvarjao:
@andrewwhitehead , @berendsliedrecht , can we get this merged for us to be able to get alpha release for the react native wrapper?

wrappers/javascript/indy-vdr-react-native/package.json
```diff
@@ -36,12 +37,10 @@
...
     "build": "yarn clean && yarn compile",
     "clean": "rimraf -rf ./build",
-    "compile": "tsc -p tsconfig.build.json",
-    "install": "node-pre-gyp install"
+    "compile": "tsc -p tsconfig.build.json"
   },
   "dependencies": {
-    "@hyperledger/indy-vdr-shared": "0.2.0-dev.6",
-    "@mapbox/node-pre-gyp": "^1.0.10"
+    "@hyperledger/indy-vdr-shared": "0.2.0-dev.6"
```
andrewwhitehead, line 43: I believe this needs updating to 0.2.2?

wrappers/javascript/indy-vdr-react-native/package.json
berendsliedrecht: > I got rid of this because the new github actions copy the compiled binaries into the native folder, so the alpha release will target the latest binaries

but it only does this in the `alpha-release` action, correct? Normal releases will not contain the copied library and fail as we also do not push them to the repository.

TimoGlastra:
> Included binaries in react native package so we have a smoother release cycle without having to wait for a binary release

It is generally not advised to push such big binaries to the NPM registry. Also because it means you download the binaries for all platforms when you install the package rather than just the binary for your platform.

So from my side this would be a No to merging this PR as is. Maybe we can look at creating pre-releases on github, so that we download the binary and not have to include it in the package.

.github/workflows/build.yml
```diff
@@ -427,6 +427,57 @@ jobs:
...
+    name: Alpha Release
+    if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+    permissions:
+      id-token: write
+    runs-on: ubuntu-latest
+    env:
+      NPM_TAG: 'alpha'
+    needs: 
+      - create-ios-xcframework
+      - create-android-library
```
TimoGlastra, line 440: This seems to just release the ios/android wrappers as alpha. Not the Node.JS.

I would really like to keep consistency between node.js and react native.

My proposal for this PR would be:
- on push to main, build the binaries for indy vdr
- update the version of the wrapper to the current package.json version but with e.g. .alpha-{commit-sha} appended
- create a prerelease on github and push the binaries
- release the JS wrapper

.github/workflows/build.yml
```diff
@@ -427,6 +427,57 @@ jobs:
...
+        with:
+          name: indy_vdr.xcframework
+          path: indy-vdr-react-native/native/mobile/ios/
+
+      # On push to main, release unstable version
+      - name: Release alpha
+        run: |
+          git update-index --assume-unchanged $(find . -type d -name node_modules -prune -o -name 'package.json' -print | tr "\n" " ")
+          #export NEXT_VERSION_BUMP=$(yarn next-version-bump)
+          export NEXT_VERSION_BUMP="major"
```
TimoGlastra, line 475: This will always increase the version by major, not sure if that's what we want?

Maybe we can just do the current version + alpha-{commit-sha}?

wrappers/javascript/indy-vdr-react-native/package.json
```diff
@@ -28,6 +28,7 @@
     "ios/indyVdr.xcodeproj/project.pbxproj",
     "cpp/**/*.cpp",
     "cpp/**/*.h",
+    "native/**",
```
TimoGlastra, line 31: This should not be published

wrappers/javascript/indy-vdr-react-native/package.json
TimoGlastra: Honestly i don't see a reason to change this setup?

It works cross platform for Node.JS / iOS / Android and means you only download the binary for your platform (in the Node.JS release)

We've put quite some thought into this process and I would be really disappointed to see it replaced by something else that caters better to your use case (react native), but doesn't take into account Node.Js

TimoGlastra (changes requested):

wrappers/javascript/indy-vdr-react-native/package.json
TimoGlastra: See my suggestion on how to approach this in the other comment

.github/workflows/build.yml
TimoGlastra: The github pre releaase could e.g. be called "JS 0.2.2 Alpha {sha-commit}" to make it clear this is a JS pre-release.

Then the package.json binary is updated to link to this pre-release.

This way pre-releases work the same as normal releases.

cvarjao: It seems messy to manage different lifecycle in the same repository. I think we need to move to [uniffi-rs](https://mozilla.github.io/uniffi-rs/latest/) or some other tool. I am concern about those wrappers getting out of sync and avoid missing release, and release from side branches, like we it was done.

PR #329, opened 2024-09-17
fix: point to hl indy-did-networks repo

dbluhm:
And a minor usability improvement for the util to enable retrieving did:indy networks from a different repo than just the hyperledger one.


dbluhm:
@swcurran quick heads up, this probably should have been done at the same time as creating the hyperledger did-indy-network repo

swcurran (approved):

PR #332, opened 2024-09-27
Loosen time dependency patch version constraint 

gmulhearn-anonyome:
Previously we were locked into time `=0.3.20`: https://github.com/hyperledger/indy-vdr/pull/166

Recently it was updated to `=0.3.36` due to issues with rust 1.80 and <0.3.36 versions of time.

However just as a crate cleanliness, it would be nice to only pin the version to 0.3.36 if we absolutely have to, dep conflicts may arise for consumers of the indy-vdr in the future as a result etc

FYI @andrewwhitehead : RE your [original decision](https://github.com/hyperledger/indy-vdr/pull/166) to lock the time version, is it still necessary?

### DIFF
--- a/libindy_vdr/Cargo.toml
+++ b/libindy_vdr/Cargo.toml
@@ -57,14 +57,13 @@ serde_json = "1.0"
 sha2 = "0.10"
 sha3 = "0.10"
 thiserror = "1.0"
-time = { version = "=0.3.36", features = ["parsing"] }
+time = { version = "0.3.36", features = ["parsing"] }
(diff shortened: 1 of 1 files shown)


andrewwhitehead:
There hasn't been a new release of the time crate past 0.3.36 so this doesn't have any effect right now. It's possible that a new release would increase the minimum Rust version past what we currently need for the Android build, but that's true of the other dependencies as well so I don't think it hurts to loosen this up for now.

andrewwhitehead (approved):

